PR #8, opened 2020-08-26
Fixing variable length integer encoding

### LINKED ISSUE #7: Variable length integers encoded incorrectly
Variable length integers appear to be encoded incorrectly.

This can cause messages with lengths larger than 128 to register as malformed packets.

ryan-summers:
This PR fixes #7 by correcting the encoding for variable length integers. The order was previously reversed.

Testing:
* An MQTT publish packet was generated with a size of 386. It was confirmed that the packet was properly digested by the broker.

### DIFF
--- a/src/ser/packet_writer.rs
+++ b/src/ser/packet_writer.rs
@@ -99,23 +99,23 @@ impl<'a> ReversedPacketWriter<'a> {
 
         if value & (0b0111_1111 << 21) > 0 {
             let data: [u8; 4] = [
-                (value >> 21) as u8 & 0x7F,
-                (value >> 14) as u8 | 0x80,
-                (value >> 7) as u8 | 0x80,
                 (value >> 0) as u8 | 0x80,
+                (value >> 7) as u8 | 0x80,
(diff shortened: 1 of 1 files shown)


src/ser/packet_writer.rs
```diff
@@ -99,23 +99,23 @@ impl<'a> ReversedPacketWriter<'a> {
...
-                (value >> 7) as u8 | 0x80,
                 (value >> 0) as u8 | 0x80,
+                (value >> 7) as u8 | 0x80,
+                (value >> 14) as u8 & 0x7F,
             ];
 
             self.write(&data)
         } else if value & (0b0111_1111 << 7) > 0 {
-            let data: [u8; 2] = [(value >> 7) as u8 & 0x7F, (value >> 0) as u8 | 0x80];
+            let data: [u8; 2] = [(value >> 0) as u8 | 0x80, ((value >> 7) & 0x7F) as u8];
```
ryan-summers, line 118: Let's add unit tests for this.

PR #15, opened 2020-09-01
Replace ad-hoc std-based NAL with std-embedded-nal

chrysn:
Granted, std-embedded-nal is quite immature yet, but hey it works :-)

(If you like this but feel held back by the git= dependency, I'm happy to upload a 0.0.1, but didn't yet as it might get some vetting in the course of this PR.)


ryan-summers:
I believe we're not able to publish to crates.io with any git= dependencies as is.

chrysn:
If that's the only issue, that is about to be mitigated.

chrysn:
There's a 0.0.1 on crates.io now; I've updated the PR accordingly, and tests here and in CI went through.

ryan-summers (approved):
:LGTM: Offloading the implementation of the stack here is definitely in our interest, as it helps to keep the focus on the test that is being performed.

Thanks for the change!

PR #23, opened 2020-10-23
Update stm32h7 Example

irwineffect:
Updates the minimq-stm32h7 example to make it easier for others to use.

Summary of changes:
- Now enables the cycle counter and icache through the init::Context rather than
taking the cortex_m::Peripherals. We previously were reliant on a bug  (rtic-rs/cortex-m-rtic#321) in
cortex_m_rtic v0.5.3 concerning being able to take the Peripherals during init.
- Removed dependency on external si7021 sensor, now publishes data based on
internal random number generator.
- Previously, the device would not reconnect to MQTT broker after restarting the
program due to smoltcp trying to reuse the previous TCP connection. This was
fixed in smoltcp-rs/smoltcp#367, but this change has
not been released on crates.io yet, so we are pulling in a specific version of
smoltcp until this is released.
- Adding the cargo lockfile to make it easier for others to reproduce a working build


examples/minimq-stm32h7/src/main.rs
```diff
@@ -213,19 +202,27 @@ const APP: () = {
             }
 
             if tick && (time % 1000) == 0 {
+                if client.is_connected().unwrap() {
```
irwineffect, line 205: I didn't intend to commit this debug code, should I remove it?

examples/minimq-stm32h7/Cargo.toml
```diff
@@ -25,7 +25,10 @@ serde-json-core = "0.1"
 path = "../../"
 
 [dependencies.smoltcp]
-version = "0.6.0"
+# Using specific git commit until PR 367 lands on crates.io
+# https://github.com/smoltcp-rs/smoltcp/pull/367
+git = "https://github.com/smoltcp-rs/smoltcp"
+rev = "bdfa442"
```
ryan-summers, line 31: You can remove the rev reference here - the `Cargo.lock` should pick up the revision and we can just follow `master` otherwise.

examples/minimq-stm32h7/Cargo.toml
```diff
@@ -13,7 +13,7 @@ panic-halt = "0.2.0"
 cortex-m-log = { version = "0.6.1", features = ["log-integration"] }
 stm32h7xx-hal = { version = "0.6.0", features = ["stm32h743v", "rt"] }
 stm32h7-ethernet = { path = "stm32h7-ethernet" }
-cortex-m-rtic = "0.5.3"
+cortex-m-rtic = "0.5.5"
```
ryan-summers, line 16: Should say `cortex-m-rtic = "^0.5.5"` I believe?

examples/minimq-stm32h7/Cargo.toml
```diff
@@ -25,7 +25,10 @@ serde-json-core = "0.1"
 path = "../../"
 
 [dependencies.smoltcp]
-version = "0.6.0"
+# Using specific git commit until PR 367 lands on crates.io
+# https://github.com/smoltcp-rs/smoltcp/pull/367
+git = "https://github.com/smoltcp-rs/smoltcp"
+rev = "bdfa442"
 default-features = false
 features = ["ethernet", "proto-ipv4", "socket-tcp", "proto-ipv6"]
```
ryan-summers, line 34: Need to remove the si7021 dependency

examples/minimq-stm32h7/src/main.rs
```diff
@@ -43,8 +42,8 @@ static mut NET_STORE: NetStorage = NetStorage {
 static mut DES_RING: ethernet::DesRing = ethernet::DesRing::new();
 
 #[derive(Serialize, Deserialize)]
-struct Temperature {
-    temperature_c: f32,
+struct Random {
+    random: Option<u32>
```
ryan-summers, line 46: Why is it an Optional param? Shouldn't it just be a u32?

examples/minimq-stm32h7/src/main.rs
```diff
@@ -119,7 +117,7 @@ const APP: () = {
         let gpioa = c.device.GPIOA.split(ccdr.peripheral.GPIOA);
         let gpiob = c.device.GPIOB.split(ccdr.peripheral.GPIOB);
         let gpioc = c.device.GPIOC.split(ccdr.peripheral.GPIOC);
-        let gpiof = c.device.GPIOF.split(ccdr.peripheral.GPIOF);
+        let _gpiof = c.device.GPIOF.split(ccdr.peripheral.GPIOF);
```
ryan-summers, line 120: Just remove the GPIO configuration that's not needed. We should ideally have this example be as simply as possible.

examples/minimq-stm32h7/src/main.rs
ryan-summers: Yeah, let's get rid of it.

examples/minimq-stm32h7/src/main.rs
```diff
@@ -213,19 +202,27 @@ const APP: () = {
...
+                    serde_json_core::to_string(&random).unwrap();
                 client
                     .publish(
-                        "temperature",
-                        &temperature.into_bytes(),
+                        "random",
+                        &random.into_bytes(),
                         QoS::AtMostOnce,
                         &[],
                     )
```
ryan-summers, line 228: Can you add a handle of `client.poll()` for `Error::Disconnected` as well? That shouldn't cause a panic - it's normal behavior when the broker disconnects.

examples/minimq-stm32h7/stm32h7-ethernet/Cargo.toml
```diff
@@ -20,7 +20,10 @@ cortex-m = "0.6.2"
 log = "0.4.8"
 
 [dependencies.smoltcp]
-version = "0.6.0"
```
ryan-summers, line 23: Honestly, I'd prefer if we just bumped the hal version of stm32h7xx-hal to the most recent and used ethernet from there - they recently got merged together.

examples/minimq-stm32h7/stm32h7-ethernet/Cargo.toml
```diff
@@ -20,7 +20,10 @@ cortex-m = "0.6.2"
...
+# Using specific git commit until PR 367 lands on crates.io
+# https://github.com/smoltcp-rs/smoltcp/pull/367
+git = "https://github.com/smoltcp-rs/smoltcp"
+rev = "bdfa442"
```
ryan-summers, line 26: The correct approach is to use [cargo patch](https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html) to avoid injecting this into all of the other toml files.

ryan-summers (changes requested):

examples/minimq-stm32h7/src/main.rs
irwineffect: The RNG returns a Result (not sure under which conditions it can fail), but the error variant of the Result wasn't serializable. I decided to just transform it into an Option that will return None in the case of failure.

examples/minimq-stm32h7/stm32h7-ethernet/Cargo.toml
irwineffect: I didn't know about cargo patch, that's a great feature!

examples/minimq-stm32h7/Cargo.toml
```diff
@@ -11,26 +11,35 @@ cortex-m-rt = "0.6.10"
...
 
-[dependencies.si7021]
-git = "https://github.com/ryan-summers/si7021-rs"
+[patch.crates-io]
+# Make dependencies use master branch for smoltcp until PR 367 lands on
+# crates.io.
+# https://github.com/smoltcp-rs/smoltcp/pull/367
+smoltcp = {git = "https://github.com/smoltcp-rs/smoltcp", branch = "master"}
```
ryan-summers, line 41: For reference, I believe `branch = "master"` is inferred, so it's not necessary to specify.

examples/minimq-stm32h7/stm32h7-ethernet/Cargo.toml
ryan-summers: Yeah, I recently discovered it as well :)

ryan-summers (approved):

PR #24, opened 2021-01-30
Updating nb result handling

### LINKED ISSUE #25: Reading the socket may block indefinitely
nb::block() is used when reading, which can cause the network stack to erroneously block indefinitely. This should likely return 0 if data would block.

ryan-summers:
This PR removes the `nb::block!()` during read requests to prevent inadvertent blocking. Instead, the read request returns 0, indicating no data read.

This also propagates a network error out of MiniMQ and removes an unwrap.

This fixes #25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This document describes the changes to Minimq between releases.
   mutation.
 * Use the `std-embedded-nal` crate as a dependency to provide a standard `NetworkStack` for
   integration tests.
+* Updating `read()` to not block when no data is available. Updating `write()` to progate network
+  stack errors out to the user.
 
 ## Version 0.1.0
 Version 0.1.0 was published on 2020-08-27
--- a/examples/minimq-stm32h7/Cargo.lock
+++ b/examples/minimq-stm32h7/Cargo.lock
@@ -456,7 +456,7 @@ dependencies = [
 [[package]]
 name = "smoltcp"
 version = "0.6.0"
-source = "git+https://github.com/smoltcp-rs/smoltcp?branch=master#bdfa44270e9c59b3095b555cdf14601f7dc27794"
+source = "git+https://github.com/smoltcp-rs/smoltcp#bdfa44270e9c59b3095b555cdf14601f7dc27794"
 dependencies = [
  "bitflags",
  "byteorder",
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -117,7 +117,11 @@ where
     fn read(&self, mut buf: &mut [u8]) -> Result<usize, Error<N::Error>> {
         let mut socket_ref = self.socket.borrow_mut();
         let mut socket = socket_ref.take().unwrap();
-        let read = nb::block!(self.network_stack.read(&mut socket, &mut buf)).unwrap();
+        let read = match self.network_stack.read(&mut socket, &mut buf) {
+            Ok(count) => count,
+            Err(nb::Error::WouldBlock) => 0,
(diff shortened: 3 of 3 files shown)


HarryMakes:
Hello, thanks for the commit! If I'm looking at it correctly, `read()` is now non-blocking and `write()` remains a blocking operation. Meanwhile, if the network stack's read or write operation returns an `nb::Error::Other`, then `read()` and `write()` would immediately return the propagated network error.

However, in the current code the socket reference is a RefCell where the socket (as an `Option::Some`) gets taken (`RefCell::take()`) before the network stack's read/write operation is performed, leaving `Option`'s default `None` in the RefCell. This means that if the network stack returns any error that isn't `nb::Error::WouldBlock`, then the original reference to the socket would be gone, and so any subsequent call of `socket_ref.take().unwrap()` (on `None`) would raise an unhandled error. In contrast, `nb::Error::WouldBlock` would ensure that `read()`/`write()` doesn't return before the socket is put back to the RefCell (`RefCell::replace()`).

In the example provided by the library, the main application loops to `poll()` the client and would handle `minimq::Error::Disconnected`, but this can only be returned if `socket_is_connected()` doesn't get stuck at the [`socket_ref.take().unwrap()`](https://github.com/quartiq/minimq/blob/6777c669604fd1c8043742314159e9090ed4f115/src/mqtt_client.rs#L247) and does call the network stack's `is_connected()`. Because of this, I imagine that if it encounters interrupted Ethernet linkup or closed connection at the broker, which would raise the network errors that aren't `nb::Error::WouldBlock`, then the main application would simply panic at `socket_ref.take().unwrap()` within the next call of `poll()`.

A possible workaround is to make the network stack's read/write function only return `nb::Error::WouldBlock`. However, this would defeat the purpose of propagating the network error, since now I won't be able (or need) to show the network error after calling `read()`/`write()`. Another possible solution is to add an `MqttClient` method that can `RefCell::replace()` the `None` in the socket reference with the original socket or a new socket, like [`connect_socket()`](https://github.com/quartiq/minimq/blob/6777c669604fd1c8043742314159e9090ed4f115/src/mqtt_client.rs#L372) but without `Option::unwrap()`, so that when there is an network error as mentioned above, the main application can re-connect the socket first before the next iteration of the looping `poll()`.

Based on my confusion, would you like to shed some lights on whether or not the current MQTT client code handles the network errors properly? If the code doesn't need modification, would you suggest some other perspectives for the main application to deal with this kind of network errors that shouldn't be fatal and should permit the client numerous retries?

HarryMakes:
> Another possible solution is to add an `MqttClient` method that can `RefCell::replace()` the `None` in the socket reference with the original socket or a new socket

Aha! I just saw that in your open PR #35, you are already proposing that [`socket_is_connected()` should do the `replace()`](https://github.com/quartiq/minimq/blob/e686ccd89090f32990661fe49416d570e2e80835/src/mqtt_client.rs#L242-L245). I guess this would eliminate all the concerns I've addressed.

However, have there been special considerations for `read()` to never block but `write()` can still block? And usually how is the user of this minimq library expected to use `nb::Error::WouldBlock` on their network stack's implementation of reads/writes? Thanks!

ryan-summers:
You are correct - it looks like there's a bug that was introduced such that the socket isn't properly replaced. In the future, please add these as issues as opposed to commenting on pull requests so they can be tracked properly. I spawned https://github.com/quartiq/minimq/issues/37 as a result of this.

HarryMakes:
Thanks a lot! I wasn't sure if this is ever an issue in the code, or just my bad or incomplete understanding of the library, so I didn't use the issue tracker. I'll probably also write a neat summary on the issue over there!

PR #31, opened 2021-03-02
Feature/optional client

### LINKED ISSUE #30: Overly strict ClientID check (?)
Per https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901059:

> The Server MUST allow ClientID’s which are between 1 and 23 UTF-8 encoded bytes in length, and that contain only the characters "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ". The Server MAY allow ClientID’s that contain more than 23 encoded bytes. The Server MAY allow ClientID’s that contain characters not included in the list given above. 

This suggests that the client ID validation check https://github.com/quartiq/minimq/blob/8e537e2a584ca0e97fc0b37994f4e7c4878c0335/src/ser/serialize.rs#L26-L34 might be better left to the server.

ryan-summers:
This PR fixes #30 by making the client ID optional. Additionally, the overly-strict client ID check has been removed to allow broker-mediated client IDs.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -85,7 +85,7 @@ where
     ///
     /// # Args
     /// * `broker` - The IP address of the broker to connect to.
-    /// * `client_id` The client ID to use for communicating with the broker. If None, rely on the
+    /// * `client_id` The client ID to use for communicating with the broker. If empty, rely on the
     ///   broker to automatically assign a client ID.
     /// * `network_stack` - The network stack to use for communication.
     ///
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -84,14 +84,15 @@ where
     ///
     /// # Args
     /// * `broker` - The IP address of the broker to connect to.
-    /// * `client_id` The client ID to use for communicating with the broker.
+    /// * `client_id` The client ID to use for communicating with the broker. If None, rely on the
+    ///   broker to automatically assign a client ID.
     /// * `network_stack` - The network stack to use for communication.
     ///
     /// # Returns
     /// An `MqttClient` that can be used for publishing messages and subscribing to topics.
-    pub fn new<'a>(
+    pub fn new(
         broker: IpAddr,
-        client_id: &'a str,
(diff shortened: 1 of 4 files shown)


src/mqtt_client.rs
```diff
@@ -84,14 +85,15 @@ where
...
+    ///   broker to automatically assign a client ID.
     /// * `network_stack` - The network stack to use for communication.
     ///
     /// # Returns
     /// An `MqttClient` that can be used for publishing messages and subscribing to topics.
-    pub fn new<'a>(
+    pub fn new(
         broker: IpAddr,
-        client_id: &'a str,
+        client_id: Option<&str>,
```
jordens, line 96: Do we need `None` here or would the empty string be indication enough? With this `Some("")` and `None` have the same meaning, right?

src/mqtt_client.rs
```diff
@@ -308,7 +315,8 @@ where
                             state.maximum_packet_size.replace(size);
                         }
                         Property::AssignedClientIdentifier(id) => {
-                            state.client_id = String::from(id);
+                            state.client_id = String::from_str(id)
+                                .or(Err(Error::ProvidedClientIdTooLong))?;
```
jordens, line 319: What happens if the broker rejects the client_id? Does it sanitize and then emit `AssignedClientIdentifier` in any case?

ryan-summers: The standard does not indicate required behavior, but it suggests that the broker should reject the message and provide a reason code. In any case, I doubt it would sanitize the ID, but it might, and if it does, we would handle that here.

src/mqtt_client.rs
ryan-summers: Removed `None` - you are correct that an empty string is the same.

src/mqtt_client.rs
jordens: ACK. Handling that case doesn't look critical to me.

jordens (approved):

PR #35, opened 2021-03-04
Updating initialization to allow unconfigured network stacks

### LINKED ISSUE #33: Support usage when network stack is not prepared during construction
If the network stack is not yet fully initialized (e.g. IP addresses are not configured due to DHCP etc.), the network stack might error during the `new()` function, which means that the client can only be constructed after the network stack is fully operational.

This isn't strictly required, and we should update `minimq` to support construction before the network stack is fully operational.

ryan-summers:
This PR fixes #33 by allowing the client to be constructed even if the network stack is not fully initialized and may generate errors.


src/mqtt_client.rs
```diff
@@ -244,12 +238,18 @@ where
...
+        if socket_ref.is_none() {
+            // Attempt to open a socket from the network stack.
+            socket_ref.replace(self.network_stack.open(Mode::NonBlocking)?);
+        }
 
+        let socket = socket_ref.take().unwrap();
+        let result = self.network_stack.is_connected(&socket);
         socket_ref.replace(socket);
 
+        let connected = result?;
```
dnadlinger, line 251: Isn't this just returning `result` in a roundabout way?

dnadlinger (approved):

dnadlinger:
Can this be squashed/merged? This solves the same issue that #36 will also have to deal with (due to keepalive disconnects).

PR #36, opened 2021-03-05
Implement keepalive (ping request/response messages)

dnadlinger:


src/mqtt_client.rs
```diff
@@ -91,27 +99,29 @@ where
...
+            socket: RefCell::new(None),
             state: RefCell::new(session_state),
             transmit_buffer: RefCell::new(GenericArray::default()),
             packet_reader: PacketReader::new(),
             connect_sent: false,
+            clock: clock,
         };
 
+        client.open_socket()?;
+        client.connect_socket()?;
```
ryan-summers, line 124: No socket operations are allowed during `new()`. If the network stack is not initialized (e.g. DHCP has not yet assigned an IP), we cannot get a socket yet and we would not be allowed to construct the client.

src/mqtt_client.rs
```diff
@@ -12,14 +12,21 @@ use embedded_nal::{nb, IpAddr, Mode, SocketAddr};
 use generic_array::{ArrayLength, GenericArray};
 use heapless::String;
 
+pub trait Clock {
+    /// Returns the current time, in seconds.
+    fn now(&self) -> u16;
+}
```
ryan-summers, line 18: Please use [`embedded-time`](https://crates.io/crates/embedded-time) instead

ryan-summers (commented):
Just some initial comments

src/mqtt_client.rs
dnadlinger: Yep, this replicates the behaviour of current master, i.e. without #35.

src/mqtt_client.rs
dnadlinger: That's a very sensible choice; I just didn't know whether there was a clear "standard" library for clocks in embedded scenarios yet, so went with the most minimal solution (a u16 seconds clock as required by MQTT). I'll make it depend on `embedded_time` instead.

dnadlinger:
Now uses embedded-time and implements connection rate limiting, but still needs testing on hardware.

dnadlinger:
This now appears to work robustly on Stabilizer using the DWT cycle counter as a clock:

```
diff --git a/src/hardware/cycle_counter.rs b/src/hardware/cycle_counter.rs
index da915c8..92be56e 100644
--- a/src/hardware/cycle_counter.rs
+++ b/src/hardware/cycle_counter.rs
@@ -1,5 +1,7 @@
 use rtic::cyccnt::{Duration, Instant, U32Ext};
 
+use core::cell::RefCell;
+use embedded_time::{clock::Error, fraction::Fraction, Clock};
 use stm32h7xx_hal::time::Hertz;
 
 /// A simple clock for counting elapsed milliseconds.
@@ -64,3 +66,28 @@ impl CycleCounter {
         self.ticks
     }
 }
+
+pub struct CycleCounterClock {
+    pub cycle_counter: RefCell<CycleCounter>,
+}
+
+impl CycleCounterClock {
+    pub fn new(counter: CycleCounter) -> CycleCounterClock {
+        CycleCounterClock {
+            cycle_counter: RefCell::new(counter),
+        }
+    }
+}
+
+impl Clock for CycleCounterClock {
+    type T = u32;
+    const SCALING_FACTOR: Fraction = Fraction::new(1, 1000);
+    fn try_now(&self) -> Result<embedded_time::Instant<Self>, Error> {
+        match self.cycle_counter.try_borrow_mut() {
+            Ok(mut c) => {
+                Ok(embedded_time::Instant::<Self>::new(c.current_ms()))
+            }
+            Err(_) => Err(Error::Unspecified),
+        }
+    }
+}
```

src/session_state.rs
```diff
@@ -2,26 +2,45 @@
...
+    C: Clock,
+{
     pub connected: bool,
-    pub keep_alive_interval: u16,
+
+    /// (embedded_time doesn't support u16.)
+    pub keep_alive_interval: Seconds<u32>,
+
+    /// Timestamp of the last message sent to the broker, for keep-alive generation.
+    pub last_write_time: Option<Instant<C>>,
```
ryan-summers, line 18: I wouldn't allow a generic message to replace the ping message. Generic messages may use QoS = 0, which means fire-and-forget. Thus, if the broker is not responding, we may continue writing and never notice via the ping mechanism. Let's just ping at a fixed rate.

src/session_state.rs
```diff
@@ -2,26 +2,45 @@
...
+
+    /// If we sent a ping request without having received a response, stores the time
+    /// the request was sent.
+    pub pending_pingreq_time: Option<Instant<C>>,
```
ryan-summers, line 22: I think this would be a lot simpler if we just kept two state variables:
1. `pending_response: bool`
2. `next_ping`: Instant<C>`

Then if we hit `next_ping` while `pending_response` is set, we disconnect from the broker. This equates to setting the ping timeout equal to the ping rate.

src/session_state.rs
```diff
@@ -2,26 +2,45 @@
...
-pub struct SessionState {
+pub struct SessionState<C>
+where
+    C: Clock,
+{
     pub connected: bool,
-    pub keep_alive_interval: u16,
+
+    /// (embedded_time doesn't support u16.)
+    pub keep_alive_interval: Seconds<u32>,
```
ryan-summers, line 15: There's really no need to bring the clock down into the session state. The `SessionState` is an MQTT abstraction and uses well-defined units for values. Let's keep the `embedded_time::Clock` out of this layer.

src/mqtt_client.rs
```diff
@@ -5,27 +5,35 @@ use crate::{
...
 use heapless::String;
 
 /// A client for interacting with an MQTT Broker.
-pub struct MqttClient<T, N>
+pub struct MqttClient<T, N, C>
 where
     T: ArrayLength<u8>,
     N: embedded_nal::TcpStack,
+    C: Clock,
+    u32: TryFrom<C::T>,
```
ryan-summers, line 22: Ideally, if the user doesn't have access to a `Clock` and doesn't want the ping request/response mechanism, they still won't be able to use this library because the constructor forces them to provide one (as well as the type signature).

We could follow the route of `smoltcp` where a generic timestamp is provided in the `update` function - any thoughts here?

src/mqtt_client.rs
```diff
@@ -5,27 +5,35 @@ use crate::{
...
 {
     /// The network stack originally provided to the client.
     pub network_stack: N,
+    pub clock: C,
```
ryan-summers, line 26: If we keep the clock, I don't think it needs to be public.

src/mqtt_client.rs
```diff
@@ -91,29 +101,29 @@ where
...
         let mut client = MqttClient {
             network_stack: network_stack,
-            socket: RefCell::new(Some(socket)),
+            socket: RefCell::new(None),
             state: RefCell::new(session_state),
             transmit_buffer: RefCell::new(GenericArray::default()),
             packet_reader: PacketReader::new(),
             connect_sent: false,
+            clock: clock,
+            ping_response_timeout: Seconds(2),
```
ryan-summers, line 123: If this is a constant, let's just make it a constant in the library. There's no need to carry it around in the `MqttClient` struct.

src/mqtt_client.rs
```diff
@@ -132,9 +142,10 @@ where
         Ok(read)
     }
 
-    fn write(&self, buf: &[u8]) -> Result<(), Error<N::Error>> {
+    fn write(&self, state: &mut SessionState<C>, buf: &[u8]) -> Result<(), Error<N::Error>> {
```
ryan-summers, line 145: This is a pretty message way of handling ownership - let's just use the session state inside of `self` and address any ownership issues resulting from that at their original source.

src/mqtt_client.rs
```diff
@@ -132,9 +142,10 @@ where
...
         let mut socket_ref = self.socket.borrow_mut();
         let mut socket = socket_ref.take().unwrap();
+        state.last_write_time.replace(self.clock.try_now().unwrap());
```
ryan-summers, line 148: If this is the only reason we need `state`, we should move `last_write_time` out of `state`.

src/mqtt_client.rs
```diff
@@ -239,12 +250,16 @@ where
...
             .map_err(|e| Error::Protocol(e))?;
-        self.write(packet)
+        self.write(&mut self.state.borrow_mut(), packet)
     }
 
     fn socket_is_connected(&self) -> Result<bool, N::Error> {
         let mut socket_ref = self.socket.borrow_mut();
-        let socket = socket_ref.take().unwrap();
+        let socket = socket_ref.take();
+        if socket.is_none() {
```
ryan-summers, line 259: This should be treated as an error condition, as it indicates we have a bug in the library.

src/mqtt_client.rs
```diff
@@ -365,27 +390,96 @@ where
...
+    }
+
     fn connect_socket(&mut self) -> Result<(), Error<N::Error>> {
+        // Don't try connecting more than once per second to avoid network spam if
+        // the broker port is closed.
+        if let Some(last_connect) = &self.last_socket_connect_time {
+            let now = self.clock.try_now().unwrap();
+            if now
+                .checked_duration_since(last_connect)
+                .and_then(|diff| Seconds::<u32>::try_from(diff).ok())
```
ryan-summers, line 415: Add a note as to why calling `ok()` on this is acceptable - this would fail and result in weird behavior if the diff were negative.

src/mqtt_client.rs
```diff
@@ -365,27 +390,96 @@ where
...
+                .map(|diff| diff < Seconds(1u32))
```
ryan-summers, line 416: Move the magic number to a constant at the top of the file along with the ping timeout.

src/mqtt_client.rs
```diff
@@ -365,27 +390,96 @@ where
...
+                .unwrap_or(false)
```
ryan-summers, line 417: Why are we unwrapping an error into `false`? Shouldn't this be expected to succeed (depending on the clock)?

src/mqtt_client.rs
```diff
@@ -365,27 +390,96 @@ where
...
+            {
+                return Err(Error::Disconnected);
+            }
+        }
+        self.last_socket_connect_time = Some(self.clock.try_now().unwrap());
```
ryan-summers, line 422: Unwrapping the clock is not safe. We need to propagate a clock error back out to the user. Ditto throughout the file.

src/mqtt_client.rs
```diff
@@ -365,27 +390,96 @@ where
...
-        // TODO: Limit the time between connect attempts to prevent network spam.
-        let socket = self
-            .network_stack
-            .connect(socket, SocketAddr::new(self.state.borrow().broker, 1883))?;
-
-        // Store the new socket for future use.
-        socket_ref.replace(socket);
+    fn is_ping_request_due(&self) -> bool {
+        let s = self.state.borrow();
+        if s.keep_alive_interval == Seconds(0_u32) {
```
ryan-summers, line 435: We should probably just wrap the `keep_alive_interval` in `Option` instead. Using an interval of 0 seconds is an old, c-like approach.

src/mqtt_client.rs
```diff
@@ -365,27 +390,96 @@ where
...
+            return false;
+        }
+        // Note(unwrap): We should have at least sent out the initial connect packet.
+        let last_write_time = &s.last_write_time.unwrap();
+        // On error (clock wrapped) be conservative and send out a keepalive ping.
+        let now = self.clock.try_now().unwrap();
```
ryan-summers, line 441: Clock errors must be returned to the user.

src/mqtt_client.rs
```diff
@@ -365,27 +390,96 @@ where
...
+        now.checked_duration_since(last_write_time)
+            .and_then(|diff| Seconds::<u32>::try_from(diff).ok())
```
ryan-summers, line 443: Document why calling `ok()` is acceptable. Ditto on the `unwrap_or` below.

src/mqtt_client.rs
```diff
@@ -447,6 +551,19 @@ where
             }
         }
 
+        if self.ping_response_is_overdue() {
```
ryan-summers, line 554: Abstract all of this logic into a `self.handle_ping()` function instead to keep this cleaner. Also, a lot of the intermediary functions that were added can likely be removed and just substituted directly into the code, since they aren't being reused.

ryan-summers (changes requested):

src/mqtt_client.rs
dnadlinger: See comment just above.

src/mqtt_client.rs
dnadlinger: Already documented three lines above.

src/mqtt_client.rs
dnadlinger: I'd argue this commit doesn't add to the mess; the mess already exists in the current version, with state split between the interior-mutable SessionState and the regular rest. I really don't want to refactor that in this PR.

src/mqtt_client.rs
dnadlinger: Ack.

src/mqtt_client.rs
dnadlinger: By the same token, the network stack wouldn't need to be public, though?

src/mqtt_client.rs
dnadlinger: Why? It's a piece of state that needs to be reset when the broker connection is reset, so it seems to belong into SessionState.

src/mqtt_client.rs
dnadlinger: I don't think so, as `open` is allowed to fail. `shutdown_socket` thus just pulls out the socket entirely without replacing it.

src/mqtt_client.rs
dnadlinger: (I will copy/paste the same comment as for the keepalive logic, though.)

src/mqtt_client.rs
dnadlinger: This comes from the existing code, and mirrors the MQTT protocol. We can always refactor that later.

dnadlinger:
I guess changing the behaviour such that ping requests are sent out regularly, no matter the other traffic, would allow us to keep the clock outside of minimq completely and just add a timestamp parameter to `poll()` instead. (The reason to have the clock was to avoid having to pass the timestamp to all different functions.)

I'll rip out the clock stuff and add a `current_seconds` parameter to poll instead, which can just be set to zero if people really don't have a clock available. (I don't want to make it optional, as inviting that much fragility should really be discouraged, and _some_ clock will be required for network handling anyway.)

src/mqtt_client.rs
ryan-summers: There's plenty of clocks that could be fallible (e.g. an external RTC over I2C)

ryan-summers:
Closing this PR, as ping support was added in #52 (thanks in large part to the work here)

PR #43, opened 2021-05-06
Feature/session state

### LINKED ISSUE #42: Add support for SessionState maintenance
Utilize session states to maintain subscriptions through MQTT disconnections.

This will simplify user-side code so that management can be offloaded to the MQTT broker.

To do this, we will need to:
1. Store the client ID assigned to us by the broker (if none was provided)
2. Request saved session state on reconnections, clean session state on new connections

Additional comments:
* Should we allow subscription requests before the broker connects and temporarily store them internally until we can send the actual request?

ryan-summers:
The PR adds support for maintaining MQTT session state. This eliminates the need to resubscribe to topics for every network disconnection.

This fixes #42 

This PR also updates ownership semantics to remove many of the RefCells used throughout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,14 @@
 
 This document describes the changes to Minimq between releases.
 
-# Unrelease
+# Unreleased
 * Client ID may now be unspecified to allow the broker to automatically assign an ID.
 * Strict client ID check removed to allow broker-validated IDs.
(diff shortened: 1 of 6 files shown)


ryan-summers:
Tested this with stabilizer and pulled the ethernet out to force an MQTT reconnect. Verified that no new subscription was sent, but that stabilizer continued responding to previous topics.

PR #44, opened 2021-05-06
Removing blocking write

### LINKED ISSUE #38: Socket writes are done in a blocking manner
If the network stack cannot accept data, the blocking write may loop indefinitely. We should instead not handle the current packet and immediately return to the user if a write cannot be completed atomically.

ryan-summers:
This PR fixes #38 by replacing the nb::block!() call with an error mapping to `Error::WriteFail`, as blocking may never complete (e.g. for network stacks like smoltcp that need software processing of ingress/egress).


ryan-summers:
Branch was tested to verify normal functionality on Stabilizer. LGTM

PR #45, opened 2021-05-27
Feature/nal update

ryan-summers:
This PR does the following:
* Update embedded-nal version to v0.6
* Remove all usage of `RefCells`
* Refactor of the API to have an `Minimq` struct and an `MqttClient`
      * this was done to resolve ownership issues with using content in the received message buffer to publish/subscribe
  * Updating to utilize const-generics.


src/mqtt_client.rs
```diff
@@ -230,46 +193,38 @@ where
...
-    fn socket_is_connected(&self) -> Result<bool, N::Error> {
-        let mut socket_ref = self.socket.borrow_mut();
-
-        if socket_ref.is_none() {
-            // Attempt to open a socket from the network stack.
-            socket_ref.replace(self.network_stack.open(Mode::NonBlocking)?);
+    fn socket_is_connected(&mut self) -> Result<bool, N::Error> {
+        if self.socket.is_none() {
+            let socket = self.network_stack.socket()?;
+            self.socket.replace(socket);
```
irwineffect, line 205: I know this is existing logic, but I wouldn't expect the name socket_is_connected() to be a fallible function with side effects and  create the socket. We could return false if the socket is none, else call network_stack.is_connected(). And then remove the Result<> from the return signature.

src/mqtt_client.rs
```diff
@@ -288,15 +243,27 @@ where
...
     }
 
-    fn handle_packet<F>(&mut self, f: &mut F) -> Result<(), Error<N::Error>>
-    where
-        for<'a> F: FnMut(&Self, &'a str, &[u8], &[Property<'a>]),
-    {
-        let packet =
-            ReceivedPacket::parse_message(&self.packet_reader).map_err(|e| Error::Protocol(e))?;
+    fn connect_socket(&mut self) -> Result<(), Error<N::Error>> {
+        let socket = self.socket.as_mut().unwrap();
```
irwineffect, line 247: Instead of unwrapping here, this could be the place that gets the socket handle.

src/mqtt_client.rs
```diff
@@ -383,21 +346,36 @@ where
...
+    /// Construct a new MQTT client.
+    ///
+    /// # Args
+    /// * `broker` - The IP address of the broker to connect to.
+    /// * `client_id` The client ID to use for communicating with the broker. If empty, rely on the
+    ///   broker to automatically assign a client ID.
+    /// * `network_stack` - The network stack to use for communication.
+    ///
+    /// # Returns
+    /// A `Minimq` interface that can be used for publishing messages and subscribing to topics.
```
irwineffect, line 361: Update comments

irwineffect (approved):

src/mqtt_client.rs
ryan-summers: The `embedded-nal::TcpClientStack::is_connected()` defines a fallible return signature, so we always have be fallible here as well. However, you make a good point about allocating the socket here. I'll see if I can find a more logical place for it.

src/mqtt_client.rs
ryan-summers: They are still mostly accurate. I've updated a few parts though - let me know what else you're referring to if I missed something

PR #46, opened 2021-07-20
Refactoring state management

### LINKED ISSUE #28: Device may not reconnect after TCP socket closed
When testing MQTT reconnection on stabilizer, I observed that stabilizer does not reconnect to the broker after the broker is restarted. The TCP connection appears to properly shut down, but needs further investigation.

image

ryan-summers:
This PR fixes #28 by refactoring state management within `minimq` to properly track various connection state details. Notably, now, after an active connection is lost, `minimq` will `close()` the current socket so that the transmit portion of the socket also closes. This was previously holding up (re)connection attempts, as the socket was still partially connected.

This PR unifies all of the connection state management into a logical sequence of steps.

TODO:
- [x] Verify state transitions are valid before accepting them.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -118,7 +118,9 @@ where
 
                 // Allocate a new socket to use and begin connecting it.
                 self.socket.replace(self.network_stack.socket()?);
-                self.connection_state.process_event(Events::GotSocket).unwrap();
+                self.connection_state
+                    .process_event(Events::GotSocket)
+                    .unwrap();
             }
 
             // In the connect transport state, we need to connect our TCP socket to the broker.
@@ -159,7 +161,9 @@ where
                 info!("Sending CONNECT");
                 self.write(packet)?;
 
-                self.connection_state.process_event(Events::SentConnect).unwrap();
+                self.connection_state
+                    .process_event(Events::SentConnect)
+                    .unwrap();
             }
 
             _ => {}
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -65,9 +65,9 @@ mod sm {
 
     statemachine! {
         transitions: {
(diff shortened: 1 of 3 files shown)


jordens:
These state machine match expressions usually become hard to read and large with lots of rightward drift. Can we use something like [smlang-rs](https://github.com/korken89/smlang-rs) to streamline this?

Also it looks like the only input to the FSM (with four states) is `socket_is_connected()` and that `process()` will advance through several states regardless of any input. That doesn't look right.

ryan-summers:
> Also it looks like the only input to the FSM (with four states) is `socket_is_connected()` and that `process()` will advance through several states regardless of any input. That doesn't look right.

This is not the case. Process only proceeds through states when no errors occur. Otherwise, the state continues retrying until successful. Thus, the complete execution of a state is also an input. Take a look at the various error unwraps (?) in the state process functions.

I've also now refactored to use `smlang`

src/mqtt_client.rs
```diff
@@ -59,6 +59,27 @@ impl<E> From<E> for Error<E> {
     }
 }
 
+mod sm {
+
+    use smlang::statemachine;
+
+    statemachine! {
+        transitions: {
+            *Restart + Update = ConnectTransport,
```
jordens, line 68: I would differentiate the two `Update` events into `GotSocket` and `SentConnect`.

jordens (changes requested):
Nice.

jordens:
> This is not the case. Process only proceeds through states when no errors occur. Otherwise, the state continues retrying until successful. Thus, the complete execution of a state is also an input. Take a look at the various error unwraps (?) in the state process functions.
> 
> I've also now refactored to use `smlang`

Ack. That refactor also makes this clear.

jordens (approved):

PR #55, opened 2021-08-20
Updating ping timeout to operate if no control packets are received

### LINKED ISSUE #53: Update pings to be generated whenever no data received from broker
After further thought on [@dnadlinger's comment on Pull #52](https://github.com/quartiq/minimq/pull/52#issuecomment-900350932), it may be prudent to generate pings whenever we haven't received a control packet from the broker as well.

Thus, pings should be generated if either of the following conditions are met:
1. We have not sent a control packet within the keep-alive window
2. The broker hasn't sent a control packet in some pre-determined period

This will ensure we're properly checking even if the packet write fails, but the packet can't be sent for some reason (e.g. async transmission via external stack like the W5500).

ryan-summers:
This PR fixes #53 by updating the ping timeout to operate if no control packets are received from the broker.

This is not required per the MQTT spec, but it should make the client more robust to disconnection events.

This PR also subtly fixes an issue with the ping timeout - previously, if no RX/TX was registered, a ping would never become due. This PR refactors the ping deadline logic to require a ping if there's no known last RX/TX time (when a deadline is set).

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -376,6 +376,8 @@ where
     where
         F: FnMut(&mut MqttClient<N, C, T>, &'a str, &[u8], &[Property<'a>]),
     {
+        self.session_state.register_reception(self.clock.try_now()?);
+
         if !self.session_state.connected {
             if let ReceivedPacket::ConnAck(acknowledge) = packet {
                 let mut result = Ok(());
--- a/src/session_state.rs
+++ b/src/session_state.rs
@@ -17,8 +17,9 @@ pub struct SessionState<C: Clock> {
     pub broker: IpAddr,
     pub maximum_packet_size: Option<u32>,
     pub client_id: String<64>,
-    last_transmission: Option<Instant<C>>,
     pub pending_subscriptions: Vec<u16, 32>,
+    last_transmission: Option<Instant<C>>,
+    last_reception: Option<Instant<C>>,
     packet_id: u16,
     active: bool,
 }
@@ -34,6 +35,7 @@ impl<C: Clock> SessionState<C> {
             packet_id: 1,
             keep_alive_interval: Some(59.seconds()),
             last_transmission: None,
+            last_reception: None,
             pending_subscriptions: Vec::new(),
             maximum_packet_size: None,
         }
@@ -46,7 +48,6 @@ impl<C: Clock> SessionState<C> {
         self.keep_alive_interval = Some(59.seconds());
         self.maximum_packet_size = None;
         self.pending_subscriptions.clear();
-        self.last_transmission = None;
     }
 
     /// Called whenever an active connection has been made with a broker.
@@ -78,12 +79,24 @@ impl<C: Clock> SessionState<C> {
         self.last_transmission = Some(now);
     }
 
+    pub fn register_reception(&mut self, now: Instant<C>) {
+        self.last_reception = Some(now);
+    }
+
     pub fn ping_is_due(&self, now: &Instant<C>) -> bool {
-        // Send a ping if we haven't sent a transmission in the last 50% of the keepalive internal.
-        self.keep_alive_interval
-            .zip(self.last_transmission)
-            .map_or(false, |(interval, last)| {
-                *now > last + 500.milliseconds() * interval.0
-            })
+        // Send a ping if we haven't sent or received data within 50% of the keep-alive interval.
+        if let Some(interval) = self.keep_alive_interval {
+            let tx_timeout = self
+                .last_transmission
+                .map_or(false, |last| *now > last + 500.milliseconds() * interval.0);
+
+            let rx_timeout = self
+                .last_reception
+                .map_or(false, |last| *now > last + 500.milliseconds() * interval.0);
+
+            tx_timeout || rx_timeout
+        } else {
+            false
+        }
     }
 }


src/session_state.rs
```diff
@@ -74,16 +75,34 @@ impl<C: Clock> SessionState<C> {
...
+    }
+
     pub fn ping_is_due(&self, now: &Instant<C>) -> bool {
-        // Send a ping if we haven't sent a transmission in the last 50% of the keepalive internal.
-        self.keep_alive_interval
-            .zip(self.last_transmission)
-            .map_or(false, |(interval, last)| {
-                *now > last + 500.milliseconds() * interval.0
-            })
+        // Send a ping if we haven't sent or received data within 50% of the keep-alive interval.
```
jordens, line 93:
```suggestion
        // Send a ping if we haven't both sent and received messages within 50% of the keep-alive interval.
```

src/session_state.rs
```diff
@@ -74,16 +75,34 @@ impl<C: Clock> SessionState<C> {
         }
     }
 
+    pub fn handle_connection_reset(&mut self) {
```
jordens, line 78: Since this is called on `States::ConnectTransport`, shouldn't `last_transmission` and `last_reception` be set to `now`?
Otherwise, on paper, if there is never anything sent and received, `ping_is_due` will not fire and the connection will stay.

OTOH: If, as you may say, you guarantee, that you send at least the connect command, then at least the last_transmission timer will fire and the ping response not coming in would ultimately trigger disconnect.

Either way, let's describe what's supposed to happen in the pathological case where both timestamps remain `None` or why that case doesn't exist.

src/session_state.rs
```diff
@@ -74,16 +75,34 @@ impl<C: Clock> SessionState<C> {
...
-        self.keep_alive_interval
-            .zip(self.last_transmission)
-            .map_or(false, |(interval, last)| {
-                *now > last + 500.milliseconds() * interval.0
-            })
+        // Send a ping if we haven't sent or received data within 50% of the keep-alive interval.
+        if let Some(interval) = self.keep_alive_interval {
+            let tx_timeout = self
+                .last_transmission
+                .map_or(false, |last| *now > last + 500.milliseconds() * interval.0);
```
jordens, line 97: Took me a while to understand `500.milliseconds() * interval.0`. What's the units of the factors and the units of the product?
`500.milliseconds() * (interval/1.seconds())` would be meaningful to me (if that is possible).
Same in L101.
And who guarantees that the comparison is done in milliseconds. Are `now` and `last` always in milliseconds?

jordens (changes requested):

src/session_state.rs
ryan-summers: This is ultimately an inconvenience imposed by the embedded-time types. Division of `Duration`s use the left-hand type, so dividing 1.seconds() / 2 = 0 seconds, which is unfortunate.

I'm going to just refactor so that we have timeouts 500ms before the deadline - the spec indicates that servers should disconnect clients at 1.5x the timeout interval.

src/session_state.rs
ryan-summers: I've refactored sot hat the last_transmission and last_reception will default to indicate timeout if no timestamp is present, which should resolve any issues here. I've also refactored state management such that these conditions should still nominally never happen, so it's a backstop against future bugs.

ryan-summers:
I have refactored after discussion in #53 so that the ping goes out regardless of current traffic. This gaurantees one ping for each keep-alive interval.

src/mqtt_client.rs
```diff
@@ -314,16 +309,32 @@ where
             &[Property<'a>],
         ),
     {
-        if !self.session_state.connected {
-            if let ReceivedPacket::ConnAck(acknowledge) = packet {
+        // All packets must be received in the active state.
+        if self.connection_state.state() != &States::Active {
```
jordens, line 313:
```suggestion
        if !self.is_connected() {
```

src/mqtt_client.rs
```diff
@@ -314,16 +309,32 @@ where
...
+                    "Received invalid packet outside of connected state: {:?}",
+                    packet
+                );
+                return Err(Error::Protocol(ProtocolError::Invalid));
+            }
+        }
+
+        match packet {
+            ReceivedPacket::ConnAck(acknowledge)
+                if self.connection_state.state() == &States::Establishing =>
```
jordens, line 328: This is the same check as in L315. I would either move the `is_connected()` check above into the other match arms as guard. Or split the ConnAck arm out of the match. Then there is a single failure arm for unsupported packet/state combination.
Overall this is a very big function. Let's try to make it a bit more readable.

src/session_state.rs
```diff
@@ -107,8 +103,15 @@ impl<Clock: embedded_time::Clock, const MSG_SIZE: usize, const MSG_COUNT: usize>
...
     /// Called whenever an active connection has been made with a broker.
-    pub fn register_connection(&mut self) {
+    pub fn register_connection(&mut self, now: Instant<Clock>) {
         self.active = true;
+        self.ping_timeout = None;
+
+        // If the keep-alive interval is specified, set up a ping to go out before the interval
+        // elapses.
+        if let Some(interval) = self.keep_alive_interval {
+            self.next_ping.replace(now + interval - 500.milliseconds());
```
jordens, line 113: I would go for half the interval (i.e. multiplicative), not some offset (additive) on it. That's how these heartbeat things are generally done and it's important that the timing scales multiplicatively with the timeout.

ryan-summers: From earlier:

> Took me a while to understand 500.milliseconds() * interval.0. What's the units of the factors and the units of the product?
> 500.milliseconds() * (interval/1.seconds()) would be meaningful to me (if that is possible).

An `embedded_time::Duration` can only be divided by integers, so it would just become `interval / 1`. There's no way to semantically "remove" the type by making it ratiometric here. The goal of `interval.0` * 500.milliseconds() is to take the internal "number" of seconds (a u32) and multiply that by 500ms. I'll write up docs around this.


> Same in L101.
> And who guarantees that the comparison is done in milliseconds. Are now and last always in milliseconds?

The comparison is not done in milliseconds. The comparison is always done in `Instant`'s. Adding a `Duration` to an `Instant` produces a new `Instant` with an offset of `Duration`.

jordens: But there is no `interval.0 * 500.milliseconds()` anymore.

`now + interval / 2` doesn't work?

ryan-summers: The correct way would be to do `(interval.0 * 1000).milliseconds() / 2`, but yeah, that should also work

PR #57, opened 2021-10-04
Add Qos 1 publish support

mladedav:
Hi, I have implemented QoS 1. The main issue is that the state now has to keep track of published messages that haven't been acknowledged yet so I used heapless similarly to subscribe with constant 16 possible pending messages (which will mean a bit over 256 * 16 = 4 kB of memory assuming 256 byte buffers for messages).

I guess this could be another const generic, but I would let you decide if you think that's something worthwhile so that there aren't too many of them.

Also one thing I was not able to do properly is resending QoS 1 messages after connect. I am not really an expert in Rust and I wasn't able to satisfy borrow-checker without cloning.
If I do:
```
                for (_id, (len, message)) in self.session_state.pending_publish.into_iter() {
                    self.write(&message[..*len])?;
                }
                ```
Then I get 
```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> src\mqtt_client.rs:449:21
    |
448 |                 for (_id, (len, message)) in self.session_state.pending_publish.into_iter() {
    |                                              ----------------------------------------------
    |                                              |
    |                                              immutable borrow occurs here
    |                                              immutable borrow later used here
449 |                     self.write(&message[..*len])?;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
    ```
    
    If you were able to help me satisfy the borrow checker without this fairly large copy operation then that would be great.


src/session_state.rs
```diff
@@ -19,12 +19,13 @@ pub struct SessionState<C: Clock> {
     pub client_id: String<64>,
     last_transmission: Option<Instant<C>>,
     pub pending_subscriptions: Vec<u16, 32>,
+    pub pending_publish: LinearMap<u16, (usize, [u8; T]), 16>,
```
ryan-summers, line 22: We'll definitely want to make this generic to allow people to tune the RAM requirements of the library to their application. Alternatively, we could restrict this to 1 or 2 messages.

ryan-summers: Also, we should make this a map of `Vec<u8, T>` to make it easy to copy messages around without having to maintain an array + len

src/mqtt_client.rs
```diff
@@ -416,6 +443,12 @@ where
                     };
                 }
 
+                // Replay QoS 1 messages
+                let to_send = self.session_state.pending_publish.clone();
```
ryan-summers, line 447: We definitely don't want to clone the entire map of pending messages. Instead, we should just iterate over the keys of the map and borrow the data out of it.

src/mqtt_client.rs
```diff
@@ -416,6 +443,12 @@ where
                     };
                 }
 
+                // Replay QoS 1 messages
```
ryan-summers, line 446: I don't think we should be sending QoS level 1 messages here, as we don't know if the connection will be accepted.

That being said, until we incorporate some kind of timer, I don't know where pending-ACK messages _should_ get resent. Perhaps we should add a `resend()` API to allow applications to explicitly request resend transmission when there's pending messages? Or we could do the resend whenever someone calls `publish()`

ryan-summers (changes requested):
Hey there, and thanks for the contribution!

There are some general discussion points I've brought up below, but in general things look like they're in-line with what I would imagine is required for QoS 1.

My main concern at this point is the amount of buffering and copying taking place, but I can't think of a great way to store pending-ACK messages without buffering them first.

ryan-summers:
> If I do:
> ```
>                 for (_id, (len, message)) in self.session_state.pending_publish.into_iter() {
>                     self.write(&message[..*len])?;
>                 }
> ```
> Then I get

I believe you're looking for just `self.session_state.pending_publish.iter()`. The [`iter()`](https://docs.rs/heapless/0.7.7/heapless/struct.LinearMap.html#method.iter) method borrows self instead of consuming it, which should satisfy ownership symantics. `into_iter()` consumes `self`, which is why the borrow-checker is complaining.

However, you may want to look at the [`values()` API](https://docs.rs/heapless/0.7.7/heapless/struct.LinearMap.html#method.values), as that just gives you a list of messages that need to be resent.

src/session_state.rs
mladedav: I don't know the preferred style here (const generics aren't in the Book so I'm not even sure where to look for it), as it's a generic I would assume short names, but `pub struct SessionState<C: Clock, const T: usize, const U: usize>` hardly seems readable, especially with prospects of adding more later.

Would `pub struct SessionState<C: Clock, const MaxMessageSize: usize, const QoS1Buffe: usizer>` be too much?

mladedav:
```
                for (len, message) in self.session_state.pending_publish.values() {
                    self.write(&message[..*len])?;
                }
```
still runs into borrow checker issues but it's somewhat new because of iterator implementation it seems
```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> src\mqtt_client.rs:449:21
    |
448 |                 for (len, message) in self.session_state.pending_publish.values() {
    |                                       -------------------------------------------
    |                                       |
    |                                       immutable borrow occurs here
    |                                       a temporary with access to the immutable borrow is created here ...
449 |                     self.write(&message[..*len])?;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
450 |                 }
    |                 - ... and the immutable borrow might be used here, when that temporary is dropped and runs the destructor for type `impl Iterator`
    ```
    
    Using `iter()` instead of `into_iter()` also has the same issues because (if I understand it correctly) we need to borrow self for the process of iteration so during the iteration we cannot do mutable borrow. If it was an array we could iterate with ordinals which actually gave me the idea of copying just the packet IDs and iterating over thos. That means we would copy at most 32 bytes assuming max 16 pending messages. There is a processing time hit there (because searching for each message is O(n) so we're at O(n^2), but I assume it's not a problem with 16 messages.
    
    That said if you were able to do it without copying, I would be interested in how you managed that.

File: src/session_state.rs (hunk 1 above)
Comment (ryan-summers) [reply src/session_state.rs:22]: Naming preference from my end would be all caps, underscored spaces. It also helps readability o put the trait bounds in a `where` clause.

E.g.
```rust
pub struct SessionState<CLOCK, const MSG_SIZE, const NUM_MESSAGES>
where
    CLOCK: Clock,
    MSG_SIZE: usize,
    NUM_MESSAGES: usize
{
}
```

src/mqtt_client.rs
mladedav: Cloning keys and iterating over those.

ryan-summers:
Ah, I see the issues you're running into with the borrow checker now. I ran into a similar issue with the packet reader, which was why I had to move it out of struct ownership. I'll take a look tomorrow at how to restructure things to simplify the ownership system, thanks for pointing this out.

src/mqtt_client.rs
mladedav: If what connection will be accepted? I thought that in this branch we've already received CONNACK and this is after all the processing of the message so I assumed we can do this here.

As for why I did this, MQTT 5 - 4.4
```
When a Client reconnects with Clean Start set to 0 and a session is present, both the Client and Server MUST resend any unacknowledged PUBLISH packets (where QoS > 0) and PUBREL packets using their original Packet Identifiers. This is the only circumstance where a Client or Server is REQUIRED to resend messages. Clients and Servers MUST NOT resend messages at any other time [MQTT-4.4.0-1]
```

When I think about it this is also the only time it makes sense to retry anything, because if everything works well the Server must send PUBACK and if there is an issue it will surely disconnect and then we retry on new connection.

I'm not against doing the `resend` API, but I'm not sure if there would be any advantage to it. Well maybe in case one of the messages is malformed and then that would cause disconnect during the replay before the user was able to do anything, but then they would not be able to resend manually either.

mladedav:
> My main concern at this point is the amount of buffering and copying taking place, but I can't think of a great way to store pending-ACK messages without buffering them first.

Ideally we could wait for PUBACK in the publish method for some short time, but we would have to go through something like `poll` to get (peek) the packets and if we got something else we would end up buffering that instead. If we don't receive too many messages though that should not be an issue and it would prevent the messages from being copied.

Techinal issue I see here is wheter we even can peek packets and in cases with bad internet (which I assume is the primary audience of this no_std package) the publish method could take longer with no added benefit.

src/session_state.rs
mladedav: It turns out const generics can't be really used in where clauses (or it's not documented and I didn't happen to ned up with the right permutation)

Also rustc complains about the casing - weirdly enough it complains about `TCP_STACK` but not `CLOCK`. So I went with what rustc wanted and changed `CLOCK` to `Clock` for consistency.

warning: type parameter `TCP_STACK` should have an upper camel case name
  --> src\mqtt_client.rs:98:19
   |
98 | pub struct Minimq<TCP_STACK, CLOCK, const MSG_SIZE: usize, const MSG_COUNT: usize>
   |                   ^^^^^^^^^ help: convert the identifier to upper camel case: `TcpStack`
   |
   = note: `#[warn(non_camel_case_types)]` on by default

src/mqtt_client.rs
ryan-summers: After reading through the MQTT specification, you're correct that this is the correct time for sending them (immediately after a reconnect ConnAck is received).

My main concern is that many network stacks have an internal buffer that may fill up, so we won't necessarily be able to send _all_ pending messages (e.g. might be able to send first 3, but not any more). We need to have a flag for which messages have been retransmitted in the current session so that we can handle sending them in the `poll()` loop some time in the future (once the network stack has transmitted data).

src/mqtt_client.rs
```diff
@@ -350,12 +366,23 @@ where
...
+        let id = self.session_state.get_packet_identifier();
+
+        let mut buffer: [u8; MSG_SIZE] = [0; MSG_SIZE];
+        let packet = serialize::publish_message(&mut buffer, topic, data, qos, id, properties)?;
+
+        self.write(packet)?;
+        self.session_state.increment_packet_identifier();
+
+        if qos == QoS::AtLeastOnce {
+            self.session_state.publish_at_least_once(id, packet);
```
ryan-summers, line 379: For republish, I think packets need to have the DUP flag set in the PUBLISH fixed header

src/ser/serialize.rs
```diff
@@ -96,11 +99,20 @@ pub fn publish_message<'a, 'b, 'c>(
...
+        packet.write_u16(id)?;
+    }
+
     packet.write_utf8_string(topic)?;
 
+    let flags :u8 = match qos {
+        QoS::AtMostOnce => 0,
+        QoS::AtLeastOnce => 2,
+        QoS::ExactlyOnce => 6,
+    };
```
ryan-summers, line 112: I would prefer we assigned QoS to be integral in value and then shift it over by 1 bit to follow the diagram for the fixed header. E.g.
```rust
use bit_field::BitField
let flags = 0u8.set_bit(1..=2, qos as u8)
                       .set_bit(3, duplicate)
```

or similar, as it then makes it less error prone and clear what's going on for flag settings

src/mqtt_client.rs
mladedav: I can't find anything except checking the return value of `send` but we cannot really stop mid-packet. Maybe we could keep some kind of state but that would have to be not just which packets were resent, but also an offset of last packet in-flight. And we would have to check before every publish that there is no packet. Otherwise we would have to block there until data at least until packet boundary can be buffered.

Otherwise we could expose some kind of `resend` API you mentioned earlier and mandate that the user has to keep calling it after reconnecting (or are reconnects hidden from users?) until it returns a value indicating success. This would allow the process to not block but this would be somewhat more difficult to use.

src/ser/serialize.rs
mladedav: I used your code for qos, but I do not want to use duplicate in this method. Instead I set it in session_state when it is saved because any retries need the DUP value set but at that point this serialize method is not called.

I hope I'm not forgetting anything else that can potentially change between resends.

```rust
        let mut buf: Vec<u8, MSG_SIZE> = Vec::from_slice(packet).unwrap();
        // Set DUP = 1 (bit 3). If this packet is ever read it's just because we want to resend it
        buf[0] |= 1 << 3;

File: src/mqtt_client.rs (hunk 4 above)
Comment (mladedav) [reply src/mqtt_client.rs:379]: Fixed as described in the comment about qos and fixed header serialization.

File: src/session_state.rs [hunk 6]
```diff
@@ -1,30 +1,34 @@
...
     pub keep_alive_interval: Option<Seconds<u32>>,
-    pub ping_timeout: Option<Instant<C>>,
+    pub ping_timeout: Option<Instant<Clock>>,
     pub broker: IpAddr,
     pub maximum_packet_size: Option<u32>,
     pub client_id: String<64>,
-    last_transmission: Option<Instant<C>>,
+    last_transmission: Option<Instant<Clock>>,
     pub pending_subscriptions: Vec<u16, 32>,
+    pub pending_publish: LinearMap<u16, Vec<u8, MSG_SIZE>, MSG_COUNT>,
```
mladedav, line 25: We should check that even if we delete messages from the middle the ordering is always based on when the message has been added to satisfy:
> When the Client re-sends any PUBLISH packets, it MUST re-send them in the order in which the original PUBLISH packets were sent (this applies to QoS 1 and QoS 2 messages)
This has also ramifications for implementation of QoS 2 since the ordering should be kept across the QoS options.

mladedav: The struct uses swap_remove for removing which I assume means putting the last value in the backing vector in the removed item's position.

The structure should be tweaked to not reorder or another container has to be found

src/mqtt_client.rs
ryan-summers: This seems complicated enough that we should just open an issue for it and fix it later. There's enough in this PR as-is. I'll open up a tracking issue and post it here.

src/session_state.rs
```diff
@@ -1,30 +1,34 @@
...
 ///
+use crate::QoS;
 use embedded_nal::IpAddr;
-use heapless::{String, Vec};
+use heapless::{String, Vec, LinearMap};
 
 use embedded_time::{
     duration::{Extensions, Seconds},
-    Clock, Instant,
+    Clock as HwClock,
```
ryan-summers, line 10: Nit: Here and throughout, let's keep `embedded_time::Clock` defined as either `Clock` or fully specified as `embedded_time::Clock`. Using `Clock as HwClock` just makes the code harder to follow for someone not familiar with the codebase in my opinion.

If we're hitting naming issues, let's make the template parameter `Clock: embedded_time::Clock`

ryan-summers (changes requested):
Getting quite close towards getting this merged! Thanks a ton for all the work you've put in

- [ ] can you add an entry into CHANGELOG.md indicating that QoS::AtLeastOnce support has been added?

src/session_state.rs
ryan-summers: Ah, I didn't realize that rustfmt prefers camel case for type parameters. If that's the case, go with whatever rustfmt says is correct. I prefer to rely on the automated tools to keep things consistent over my own personal preference. Thanks for pointing this out to me!

ryan-summers:
FYI: https://github.com/quartiq/minimq/pull/58 refactors the code to remove the dependency that is forcing us to do cloning of the pending messages map. We can resolve the copy issue once that PR gets merged.

mladedav:
I've played around with iterators in the session state and I tried to integrate it with #58 this but I don't think this is enough for the issue here.
In the code as follows `session_state` is borrowed to iterate through the pending publish packets (directly or indirectly), but then it cannot be borrowed mutably to be passed to `write_packet`.

                for packet in self.session_state.iterate_pending_publish() {
                    self.network
                        .write_packet(&mut self.session_state, &self.clock, packet)?
                }

That said I like the separation as it's done here, it just doesn't seem to fix the issue with borrowing in the other PR. For that I believe we would have to decouple the pending publishes and ping timer in the session state so that they can be borrowed independently.

ryan-summers:
> That said I like the separation as it's done here, it just doesn't seem to fix the issue with borrowing in the other PR. For that I believe we would have to decouple the pending publishes and ping timer in the session state so that they can be borrowed independently.

That's coming in https://github.com/quartiq/minimq/pull/55, so a combination of #55 and #58 should resolve this.

mladedav:
Cool, you're way ahead of me 👍

ryan-summers (commented):
While everything looks great, can you please run `cargo fmt` on your code to satisfy the style checker?

It also looks like the integration test is having problems. Other than that, I think this PR is ready

src/mqtt_client.rs
```diff
@@ -436,6 +486,13 @@ where
                 Ok(())
             }
 
+            ReceivedPacket::PubAck(ack) => {
+                // No matter the status code the message is considered acknowledged at this point
+                self.session_state.handle_puback(ack.packet_identifier);
```
ryan-summers, line 491: We need to look at the ack ReasonCode (if present) as well to verify that the server gave us a good ACK

ryan-summers: But doesn't that imply that we still need to only handle the puback if reason_code >= 0x80 || reason_code == 0x00?

E.g. if server responds with reason_code 0x10, we shouldn't actually register the packet as sent, since there's no subscriber to actaully accept the message and we can't guarantee AtLeaseOnce

ryan-summers: Hmmm... I think I can get on-board with ignoring the reason code. I'm unsure what the corrective action would even be if we got an invalid reason code (hold onto it forever?). Let's move forward with ignoring the reason code for now then. If we need to, we can add logging of it in the future.

ryan-summers:
Looks like just a new run of `cargo fmt --all` and this is good to be merged.

src/session_state.rs
```diff
@@ -46,9 +54,58 @@ impl<C: Clock> SessionState<C> {
...
+            }
+        }
+        self.pending_publish_ordering.pop();
+    }
+
+    /// Indicates if publish with QoS 1 is possible.
+    pub fn can_publish(&self, qos: QoS) -> bool {
+        match qos {
+            QoS::AtMostOnce => true,
+            QoS::AtLeastOnce => self.pending_publish.len() < MSG_SIZE,
```
ryan-summers, line 96: Should be compared against MSG_COUNT

mladedav: Right, good eye

PR #58, opened 2021-10-05
Refactoring network stack management into holder class

ryan-summers:
This PR refactors the network stack / network socket management into a separate structure. This addresses borrow-checker/ownership semantic issues that arise when trying to write control packets that are buffered internally in minimq (e.g. QoS > 0). Specifically, this is resolved because `MqttClient` no longer needs to be borrowed to write the packet, as you can now borrow the `InterfaceHolder` directly for writes.

This addresses issues identified in https://github.com/quartiq/minimq/pull/57

CC @mladedav

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This document describes the changes to Minimq between releases.
 keep-alive interval
 * Updating tests to use `std-embedded-time`
 * Fixing main docs.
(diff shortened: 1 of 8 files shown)


SingularitySurfer (commented):
I couldn't find any actual code issues.

SingularitySurfer (approved):

PR #67, opened 2021-12-06
Release/0.5.0

ryan-summers:
This PR prepares for release 0.5.0, which supports Will messages.


jordens (changes requested):
Could we add links to the diffs for each release and `unreleased`? That makes both checking the changelog when preparing a release and understanding it when looking at a release much easier.
Like https://raw.githubusercontent.com/quartiq/idsp/main/CHANGELOG.md or others.

jordens (approved):

PR #68, opened 2021-12-07
Implement finishing partial writes

mladedav:
I've been running into issues with getting RST packets from brokers fairly often if there was high load. This is painful especially because starting the connection takes really long time for some reason. It turns out my TCP stack just couldn't keep up and in the current implementation the `WriteFail` error is returned if the write would either block or if it was partially written. The latter usually breaks the MQTT connection because there is currently no way to finish the packet send.

The issue is with the partial writes which usually contained the header including length of the MQTT message but only the first half was actually transferred, the rest was silently dropped. On a subsequent write the first half of the next packet would be interpreted by the broker as the second half of the preceding packet and the rest as a new MQTT packet, which usually turned out to be malformed so the broker sent a disconnect packet and after some time closed the TCP connection.

This solution is fairly simple although it eats up a bit more space (because the network manager must be able to store up to a full packet). I wasn't very successful with convincing the borrow checker that what I want to do is safe so the unfinished packet ends up being copied once after the unsuccessful send and once when we want to try to finish.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -99,6 +99,7 @@ pub enum ProtocolError {
 pub enum Error<E> {
     Network(E),
     WriteFail,
+    PartialWrite,
     NotReady,
     Unsupported,
     ProvidedClientIdTooLong,
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -68,7 +68,7 @@ pub struct MqttClient<
     const MSG_SIZE: usize,
     const MSG_COUNT: usize,
 > {
-    pub(crate) network: InterfaceHolder<TcpStack>,
+    pub(crate) network: InterfaceHolder<TcpStack, MSG_SIZE>,
     clock: Clock,
     session_state: SessionState<Clock, MSG_SIZE, MSG_COUNT>,
     connection_state: StateMachine<Context>,
@@ -283,6 +283,16 @@ where
         }
 
         Ok(())
+    }
+
+    /// Finish sending last packet if the previous operation returned Error::PartialWrite
+    ///
+    /// # Note
+    /// If this function is not called after partial write until the packet is
+    /// fully sent, the packets will most likely seem malformed to the receiving
+    /// party and they will close the connection.
+    pub fn finish_write(&mut self) -> Result<(), Error<TcpStack::Error>> {
+        self.network.finish_write()
     }
 
     fn handle_connection_acknowledge(
--- a/src/network_manager.rs
+++ b/src/network_manager.rs
@@ -5,17 +5,21 @@
 //! simple ownership semantics of reading and writing to the network stack. This allows the network
 //! stack to be used to transmit buffers that may be stored internally in other structs without
 //! violating Rust's borrow rules.
+use core::mem;
+
 use embedded_nal::{nb, SocketAddr, TcpClientStack};
+use heapless::Vec;
 
 use crate::Error;
 
 /// Simple structure for maintaining state of the network connection.
-pub(crate) struct InterfaceHolder<TcpStack: TcpClientStack> {
+pub(crate) struct InterfaceHolder<TcpStack: TcpClientStack, const MSG_SIZE: usize> {
     socket: Option<TcpStack::TcpSocket>,
     network_stack: TcpStack,
+    unfinished_packet: Option<Vec<u8, MSG_SIZE>>,
 }
 
-impl<TcpStack> InterfaceHolder<TcpStack>
+impl<TcpStack, const MSG_SIZE: usize> InterfaceHolder<TcpStack, MSG_SIZE>
 where
     TcpStack: TcpClientStack,
 {
@@ -24,6 +28,7 @@ where
         Self {
             socket: None,
             network_stack: stack,
+            unfinished_packet: None,
         }
     }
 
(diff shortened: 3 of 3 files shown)


ryan-summers:
I'm not the biggest fan of this implementation because it forces everyone to carry around an additional buffer for one full message size (which can be rather large). What are you using for your underlying `TcpStack` implementation? It seems to me like you could buffer there instead, and then have some method to flush the output buffers that you call periodically (similar to `smoltcp::Interface::poll()`.

You would then use the `SharedStack` to have access to multiple references to the underlying stack.

mladedav:
I'm using `std_embedded_nal::Stack` (but on a constrained device, I just happen to be able to also use std). I haven't found anything that would let me check or flush the buffers so I kind of assumed this would have to be solved here somehow.

Or do you maybe it's something that should actually be incorporated into `embedded_nal`? I mean flushing and at least information about buffers, available space/bandwidth, etc.?

I am not sure about the implementation of the tcp stacks, but I assumed that the partial write error means that their internal buffer itself was full and couldn't hold the whole packet. I kind of expect them to already buffer some data, although I don't have anything concrete to base that off of.

mladedav:
I also tried to come up with any option that wouldn't have to buffer the message anywhere, but we either need to save the packet, or we need to be able to recreate it which could be pretty hard considering it could originate almost anywhere like `publish` `subscribe` or even during connection to the broker in an extreme edge case.

src/network_manager.rs
```diff
@@ -5,17 +5,21 @@
...
+use heapless::Vec;
 
 use crate::Error;
 
 /// Simple structure for maintaining state of the network connection.
-pub(crate) struct InterfaceHolder<TcpStack: TcpClientStack> {
+pub(crate) struct InterfaceHolder<TcpStack: TcpClientStack, const MSG_SIZE: usize> {
     socket: Option<TcpStack::TcpSocket>,
     network_stack: TcpStack,
+    unfinished_packet: Option<Vec<u8, MSG_SIZE>>,
```
ryan-summers, line 19: Instead of buffering the message here, why don't we leverage the `pending_publish()` vector in `SessionState`? We already have data there, and oculd potentially replace the LinearMap with something like a `heapless::Pool` of messages. Then, we can just allocate a buffer from the pool for storing a pending unfinished packet.

This has the benefit of letting the user tailor the buffering needs to the RAM avialability of their target.

ryan-summers (changes requested):

src/network_manager.rs
ryan-summers: After looking into this more, I'm more convinced this approach that you have is correct. The session state buffers are for QoS messages, and mixing in a partially written message with those messages doesn't make sense to me.

src/network_manager.rs
```diff
@@ -88,13 +93,24 @@ where
...
                 } else {
                     Ok(())
                 }
             })
     }
 
+    /// Finish writing an MQTT control packet to the interface if the last write returned Error::PartialWrite
+    ///
+    pub fn finish_write(&mut self) -> Result<(), Error<TcpStack::Error>> {
+        let packet = mem::replace(&mut self.unfinished_packet, None);
```
ryan-summers, line 107:
```rust
if let Some(packet) = self.unfinished_packet.take() {
    self.write(packet.as_slice())?;
}

Ok(())
```

src/mqtt_client.rs
```diff
@@ -285,6 +285,16 @@ where
         Ok(())
     }
 
+    /// Finish sending last packet if the previous operation returned Error::PartialWrite
+    ///
+    /// # Note
+    /// If this function is not called after partial write until the packet is
+    /// fully sent, the packets will most likely seem malformed to the receiving
+    /// party and they will close the connection.
+    pub fn finish_write(&mut self) -> Result<(), Error<TcpStack::Error>> {
```
ryan-summers, line 294: Let's add a call to `client.finish_write()` on every execution of the `Minimq.poll()` function to flush these messages automatically. There's no need for the user to manually do this.

src/network_manager.rs
```diff
@@ -88,13 +93,24 @@ where
             })
             .and_then(|written| {
                 if written != packet.len() {
-                    Err(Error::WriteFail)
+                    self.unfinished_packet = Some(Vec::from_slice(&packet[written..]).unwrap());
```
ryan-summers, line 96: I don't think the user actually cares if a partial write occurs. Ultimately, we'll come back around and finish the write as soon as we have space in the network stack. Just return `Ok(())` here, since we'll finish the write later.

src/lib.rs
```diff
@@ -99,6 +99,7 @@ pub enum ProtocolError {
 pub enum Error<E> {
     Network(E),
     WriteFail,
+    PartialWrite,
```
ryan-summers, line 102: Remove this error variant

src/network_manager.rs
ryan-summers:
```rust
self.unfinished_packet.replace(Vec::from_slice(&packet[written..]).unwrap());
```

ryan-summers (changes requested):
We should also modify `can_publish()` to check if there's a pending-write.

We need to explicitly disable writing a packet if there's a partial write available, or we will corrupt the MQTT stream.

PR #75, opened 2021-12-10
Handling partial MQTT packet writes

### LINKED ISSUE #74: Partially transmitted packets cause broker disconnection
As pointed out in https://github.com/quartiq/minimq/pull/68, any partial packet write ultimately results in a broker disconnection because the broker fails to receive a complete MQTT packet.

Currently, there's no mechanism to check and see if a packet can be entirely written before we write it, so we likely need to do the following:
* Buffer any yet-to-be-written packet data
* Provide an API to check to see if Minimq can send a packet (e.g. the buffer is not in use)

ryan-summers:
This PR fixes #74 by incorporating changes from #68 with additional safeguards.

Specifically, any attempt to write a packet is now safe guarded by checking that packet writes are acceptable. It's invalid to attempt to send a packet while a partial packet is still being transmitted, as this would corrupt the MQTT connection.

CC @mladedav

### DIFF
--- a/src/network_manager.rs
+++ b/src/network_manager.rs
@@ -14,7 +14,7 @@ use crate::Error;
 pub(crate) struct InterfaceHolder<TcpStack: TcpClientStack, const MSG_SIZE: usize> {
     socket: Option<TcpStack::TcpSocket>,
     network_stack: TcpStack,
-    unfinished_packet: Option<Vec<u8, MSG_SIZE>>,
+    pending_write: Option<Vec<u8, MSG_SIZE>>,
 }
 
 impl<TcpStack, const MSG_SIZE: usize> InterfaceHolder<TcpStack, MSG_SIZE>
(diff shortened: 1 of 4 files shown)


src/network_manager.rs
```diff
@@ -24,9 +26,15 @@ where
         Self {
             socket: None,
             network_stack: stack,
+            unfinished_packet: None,
```
jordens, line 29: `pending_write`? It's not a packet.

src/network_manager.rs
```diff
@@ -79,6 +91,9 @@ where
     /// # Args
     /// * `packet` - The packet to write.
     pub fn write(&mut self, packet: &[u8]) -> Result<(), Error<TcpStack::Error>> {
```
jordens, line 93: `message` (as used for the same thing in other places) or `data` would be better since there is no guaranteed correlation with (Ethernet/TCP) packets.

jordens (changes requested):

jordens (approved):

ryan-summers:
@mladedav Thanks for pointing out the other issue. I've added it to the issue tracker and submitted a PR.

In the future, I'd recommend posting things like this as an issue - I nearly missed this comment

PR #78, opened 2021-12-14
Preparing for 0.5.2

ryan-summers:
This PR prepares for release of minimq 0.5.2, which includes patches so that partial packet writes don't break the broker connection.


jordens (approved):
:+1:

PR #79, opened 2021-12-15
Example: Add explanation what MSG_COUNT size means.

systec-ms:


ryan-summers (approved):
I noticed this last go-around at the docs, but forgot to fix it. Thanks for the PR!

PR #81, opened 2022-02-14
Deriving PartialEq for properties

ryan-summers:
This PR fixes the `Property` enum to derive PartialEq to allow easy comparison of properties. This is useful e.g. for comparing CorrelationData.


jordens (approved):
Yes.
https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits

PR #85, opened 2022-04-28
Allow configuration of non default port

brummer-simon:
This PR adds the possibility to set custom target port numbers deviating from 1883.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -7,7 +7,7 @@ use crate::{
     ser::serialize,
     session_state::SessionState,
     will::Will,
-    Error, Property, ProtocolError, QoS, Retain, {debug, error, info},
+    Error, Property, ProtocolError, QoS, Retain, MQTT_INSECURE_DEFAULT_PORT, {debug, error, info},
 };
 
 use embedded_nal::{IpAddr, SocketAddr, TcpClientStack};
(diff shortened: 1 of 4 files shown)


ryan-summers (changes requested):
Hey, thanks for the PR! This looks wonderful - I hadn't considered using TLS with minimq since none of our network stacks support it, but indeed this library is agnostic to the underlying network stack and could well be used on networks that support TLS.

Can you please add an entry for this in the CHANGELOG.md file?

Other than that, I'd like to add one run-time check when trying to modify the broker port to prevent user foot-guns :)

src/mqtt_client.rs
```diff
@@ -188,6 +187,17 @@ where
...
 
+    /// Set custom MQTT port to connect to.
+    ///
+    /// # Note
+    /// This must be completed before connecting to a broker.
+    ///
+    /// # Args
+    /// * `port` - The Port number to connect to.
+    pub fn set_broker_port(&mut self, port: u16) {
+        self.session_state.broker.set_port(port);
```
ryan-summers, line 198:
```suggestion
        if (*self.connection_state.state() != States::Restart) 
            return Err(Error::NotReady)'
        }
        
        self.session_state.broker.set_port(port);
```

brummer-simon:
@ryan-summers: I'll add the changes. We don't have a proven Networkstack with TLS either, we try to piggy-back TLS ontop of a embedded_hal::TcpClientStack implementation. We will see if this works.

src/mqtt_client.rs
ryan-summers: Correct - in the `Restart` state, we don't have any socket allocated yet from the stack. After we get it, we transition to `ConnectTransport` where we call the `TcpClientStack::connect()` command. If we just check for the `Restart` state, we can ensure that we haven't called `connect()` yet.

brummer-simon:
@ryan-summers: All requested changes were added.

ryan-summers (approved):

PR #88, opened 2022-08-17
Refactoring MQTT state management

ryan-summers:
This PR refactors the state management of the client to be entirely wrapped up in an `smlang::StateMachine`.

This change is intended to support more complex state management that will be required to support all QoS levels.

This PR also removes public access to the `Minimq.client` object. It is now exposed via `Minimq::client()` to support compatibility going forward.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -269,7 +269,6 @@ impl<
         self.network.write(packet).and_then(|_| {
             info!("Subscribing to `{}`: {}", topic, packet_id);
 
-            // TODO: Generate event?
             self.sm
                 .context_mut()
                 .session_state
@@ -384,7 +383,6 @@ impl<
             .increment_packet_identifier();
 
         if qos == QoS::AtLeastOnce {
-            // TODO: Generate an event to update session state.
             self.sm
                 .context_mut()
                 .session_state
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -302,7 +302,7 @@ impl<
     /// # Returns
     /// True if the client is connected to the broker.
     pub fn is_connected(&mut self) -> bool {
-        [&States::Active].contains(&self.sm.state())
+        matches!(self.sm.state(), &States::Active)
     }
 
     /// Get the count of unacknowledged messages at the requested QoS.
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -25,16 +25,13 @@ mod sm {
         transitions: {
             *Disconnected + Reallocated = Restart,
             Restart + SentConnect = Establishing,
-            Establishing + Connected(ConnAck<'a>) [ handle_connack ] / start_republish = Republish,
-
-            Republish + RepublishComplete = Active,
+            Establishing + Connected(ConnAck<'a>) [ handle_connack ] = Active,
 
             Active + SendTimeout = Disconnected,
             _ + ProtocolError = Disconnected,
 
             Active + ControlPacket(ReceivedPacket<'a>) [handle_packet] = Active,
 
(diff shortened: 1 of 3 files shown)


src/mqtt_client.rs
```diff
@@ -216,53 +186,63 @@ where
...
+        network.write(packet).and_then(|_| {
             info!("Subscribing to `{}`: {}", topic, packet_id);
-            self.session_state
+            session_state
                 .pending_subscriptions
                 .push(packet_id)
                 .map_err(|_| Error::Unsupported)?;
-            self.session_state.increment_packet_identifier();
+            session_state.increment_packet_identifier();
             Ok(())
```
jordens, line 216: I'd push this code down into `ClientContext`.

src/mqtt_client.rs
```diff
@@ -315,83 +295,117 @@ where
...
-        self.session_state.increment_packet_identifier();
+        network.write(packet)?;
+        session_state.increment_packet_identifier();
 
         if qos == QoS::AtLeastOnce {
-            self.session_state.handle_publish(qos, id, packet);
+            session_state.handle_publish(qos, id, packet);
+        }
+
+        Ok(())
```
jordens, line 318: Likewise should move down into `ClientContext`.

src/mqtt_client.rs
```diff
@@ -315,83 +295,117 @@ where
...
+            let message = session_state.pending_publish.get(key).unwrap();
+            network.write(message)?;
         }
 
-        result
+        if session_reset {
+            Err(Error::SessionReset)
+        } else {
+            Ok(())
+        }
```
jordens, line 408: Likewise down into `ClientContext`.

src/mqtt_client.rs
ryan-summers: I don't think we can push this all the way down without introducing a whole ton of boilerplate code.

Specifically, we need:
1. To verify that `subscribe()` occurs after we are connected, since we can't subscribe before we're connected via MQTT.
2. Provide the `subscribe()` mechanism to the `MqttClient` object, so that it can be called in both `minimq.client().poll()` and in `minimq.client()`.

If we pushed this code down into `ClientContext`, we'd have to expose a new `subscribe` function on the `MqttClient` that checked the state and then just called into the `ClientContext::subscribe()` function with all of the duplicated arguments.

We can do that, it just felt like a lot of copy-paste boilerplate to me.

src/mqtt_client.rs
ryan-summers: See above re: subscribe, as we have the same issue here

src/mqtt_client.rs
jordens: Yes, there would be a `ClientContext::subscribe()` and you would call that from a two-line function `MqttClient::subscribe()`. But that's instead of the all the destructuring of `context_mut()` in `MqttClient`.
My goal is to contain `network` and `session_state` code within the context that owns them and not leak their management code trough `pub(crate) network` and destructuring.

ryan-summers: Yeah, I'm in agreement and in the process of cleaning some of it up.

jordens: For cases where we don't need `state()` or `process_event()`, maybe a reasonable compromise is to just make it `MqttClient::context_mut().subscribe()`.

ryan-summers:
@jordens I elected to restructure it so that we only keep the `SessionState` as the state machine context. This should still facilitate easy management of the session state structure. There's still a small amount of reaching into the internal context, but I want to try to fix it up a bit later.

jordens (changes requested):
Just a minor thing. LGTM otherwise

jordens (approved):

Issue #80, opened 2022-02-03
Result codes from poll closure are not possible


ryan-summers:
The closure in the `poll()` function cannot be propagated out of the stack. Sometimes, it's desireable to have a result move from the internal closure to external context to handle control flow.

Difficulties:
* The `poll()` closure may execute multiple times, so if some `Ok(x)` were returned, only the last value could be provided

Perhaps we could allow an `Error` to exit the `poll()` loop early to allow `Error` codes to propagate outwards, but this wouldn't allow external `Ok(result)` data to be propagated.

PR #95, opened 2022-08-23
Updating parsing to avoid copying

ryan-summers:
This PR refactors the packet reception mechanism in Minimq to receive directly from the network stack into the `PacketReader`. The `PacketParser` has been broken off the `PacketReader` in anticipation of deserialization refactoring in the future.

This avoids one copy of the MQTT packet from the receive buffer into the packet reader.

Additionally, this PR updates it so that only a single MQTT packet is read from the network per `poll()` execution at most, which paves the way for fixing #80.

### DIFF
--- a/src/de/deserialize.rs
+++ b/src/de/deserialize.rs
@@ -98,14 +98,14 @@ pub enum ReceivedPacket<'a> {
 }
 
 impl<'a> ReceivedPacket<'a> {
-    /// Parse a message out of a `PacketParser` into a validated MQTT control message.
+    /// Parse a validated MQTT control packet out of a `PacketParser`.
     ///
     /// # Args
     /// * `packet_reader` - The reader to parse the message out of.
     ///
     /// # Returns
     /// A packet describing the received content.
-    pub(crate) fn parse_message<'reader: 'a>(
+    pub(crate) fn parse_packet<'reader: 'a>(
         packet_reader: &'reader PacketParser<'_>,
     ) -> Result<ReceivedPacket<'a>, Error> {
         let (message_type, flags, remaining_length) = packet_reader.read_fixed_header()?;
@@ -298,7 +298,7 @@ mod test {
         ];
 
         let reader = PacketParser::new(&serialized_connack);
-        let connack = ReceivedPacket::parse_message(&reader).unwrap();
+        let connack = ReceivedPacket::parse_packet(&reader).unwrap();
         match connack {
             ReceivedPacket::ConnAck(conn_ack) => {
                 assert_eq!(conn_ack.reason_code, 0);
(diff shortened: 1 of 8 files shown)


src/de/deserialize.rs
```diff
@@ -312,17 +309,17 @@ mod test {
 
     #[test]
     fn deserialize_good_publish() {
-        let mut serialized_publish: [u8; 7] = [
+        let serialized_publish: [u8; 7] = [
             0x30, // Publish, no QoS
-            0x04, // Remaining length
```
ryan-summers, line 317: I'm somewhat unsure why this was passing in the past. It looks completely wrong to me - there's obviously 5 bytes following the packet.

src/de/packet_parser.rs
```diff
@@ -0,0 +1,133 @@
+use crate::{warn, MessageType, Property, ProtocolError as Error};
+use bit_field::BitField;
+use heapless::Vec;
+
+pub(crate) struct PacketParser<'a> {
+    pub buffer: &'a [u8],
+    index: core::cell::RefCell<usize>,
```
jordens, line 7: I wonder what the original reason for the `RefCell` was, whether we can get rid of it now, and instead make the methods require `&mut self`.

jordens (changes requested):
Very nice cleanup.

src/de/packet_parser.rs
ryan-summers: The reason is because we are borrowing data out of `buffer` to create `Property` objects. As a result of the immutable borrow for the `buffer`, `self` can no longer mutate. Hence, why `index` is stored in a `RefCell`.

I tried removing it during this refactor, but discovered that issue.

src/de/deserialize.rs
```diff
@@ -96,15 +98,15 @@ pub enum ReceivedPacket<'a> {
 }
 
 impl<'a> ReceivedPacket<'a> {
-    /// Parse a message out of a `PacketReader` into a validated MQTT control message.
+    /// Parse a message out of a `PacketParser` into a validated MQTT control message.
```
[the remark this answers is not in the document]
ryan-summers: The MQTT specification actually uses them interchangeably (e.g. `MessageExpiryInterval` is a property, but the data structure is assigned a `PacketIdentifier`).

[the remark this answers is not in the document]
ryan-summers: Upon further review, MQTT defines "Application Message" as the contents of a publication (e.g. topic, QoS, properties, and payload).

A "Control Packet" is any of the formally defined MQTT packet types, such as PUBLISH, CONNECT, CONNACK, etc.

So it seems that PUBLISH is a "Control Packet" that also contains an "Application Message". Thus, the terminology here is correct. The `PacketParser` parses "Control Packets". 

I think the only API change this necessitates is renaming the `serialize::*_message` functions to `serialize::*_packet`.

src/de/packet_parser.rs
jordens: Ack

src/de/deserialize.rs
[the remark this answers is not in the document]
jordens: Meh. Ok.

jordens (approved):

PR #99, opened 2022-08-23
Feature/doc cleanup

### LINKED ISSUE #56: Connecting and publishing
Hi, I really like the library and I'm trying to use the master version (which I noticed is considerably different from 0.3 which is latest), but I can't seem to be able to publish. The problem seems to be in that the connection is never established, so all the publish calls are silently dropped (might I suggest logging that?) but I have not seen any API for connecting.

Can you point me to the correct function to establish a connection so that I can publish messages to a broker?

ryan-summers:
This PR fixes #56 by updating the documents to be clear that `poll()` needs to be called repeatedly to update the MQTT state.

It also cleans up documentation in various places to keep it all up-to-date.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -161,6 +161,7 @@ impl<T> From<sm::Error<Error<T>>> for Error<T> {
     }
 }
 
+/// The client that can be used for interacting with the MQTT broker.
 pub struct MqttClient<
     TcpStack: TcpClientStack,
     Clock: embedded_time::Clock,
@@ -253,6 +254,9 @@ impl<
     /// A subscription is not maintained across a disconnection with the broker. In the case of MQTT
     /// disconnections, topics will need to be subscribed to again.
     ///
+    /// The subscription will not be completed immediately. Call
+    /// `MqttClient::subscriptions_pending()` to check for subscriptions being completed.
+    ///
     /// # Args
     /// * `topic` - The topic to subscribe to.
     /// * `properties` - A list of properties to attach to the subscription request. May be empty.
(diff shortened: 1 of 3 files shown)


src/mqtt_client.rs
```diff
@@ -546,6 +548,10 @@ impl<
 }
 
 /// The general structure for managing MQTT via Minimq.
+///
+/// # Note
+/// To connect and maintain an MQTT connection, the `Minimq::poll()` method must be repeatedly
+/// called.
```
jordens, line 554:
```suggestion
/// To connect and maintain an MQTT connection, the `Minimq::poll()` method must be
/// called regularly. It should also be called repeatedly until it returns `Ok(None)` (or `Err()`).
```

jordens (changes requested):
LGTM.

src/mqtt_client.rs
ryan-summers: I don't think that update is accurate - `Ok(None)` will also be returned when processing a control packet (e.g. CONNACK), but doesn't necessarily mean there's no packets left.

jordens: Ack. But then we should make the comment true by improving the code.
Can we make `poll()` process packets until `client.handle_packet()` returns `Ok(Some())`, i.e. until it did call `f()`?

ryan-summers: I've updated the docs on `poll()` and rephrased this to indicate that `poll()` must be called regularly.

ryan-summers: That's not a bad idea, but I'd say it's out of scope here. Will open an issue for it.

ryan-summers: I spawned https://github.com/quartiq/minimq/issues/102 to handle this

jordens (approved):

PR #105, opened 2022-09-05
Refactoring `poll()` usage

### LINKED ISSUE #102: Update `poll()` behavior
Update `poll()` such that it continually processes MQTT control packets. Specifically, it should continue processing until one of the following conditions are met:
* A full packet could not be read from the network
* A single `Publish` packet has been processed

This will ensure that:
1. The client processes all ingressed data (as a best effort).
2. Results from the `poll()` closure can be returned.

ryan-summers:
This PR updates the functionality of `poll()`. Users are now informed that they should call it repeatedly until it returns `Ok(None)`. Functionality has changed such that `poll()` will process as many MQTT control packets as possible until a `publish` message is received.

This PR fixes #102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ the `poll()` closure executes on an inbound `Publish` message.
 * [breaking] External crate is now used for `varint` encoding. Varints changed to u32 format.
 * Deserialization and serialization is now handled directly by `serde`.
 * [breaking] Properties are now wrapped in MQTT-specific data types.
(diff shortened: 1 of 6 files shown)


jordens (approved):
LGTM.
Did a quick check of the tests, and couldn't find anything where it would be strictly better to poll repeatedly before doing other things. Maybe https://github.com/quartiq/minimq/blob/297e24e5526315a814fa838b4880736010f5757e/tests/poll_result.rs#L22-L29 though?

PR #107, opened 2022-09-05
Adding support for simply replying to inbound messages

### LINKED ISSUE #84: [Proposal] Add a `reply` API
Often for bi-directional MQTT traffic, some message is generated as a response to some in-bound traffic.

This response generally wants to:
1. Copy CorrelationData from the original message into the response property to correlate the two message
2. Use a provided topic in a `ResponseTopic` as the destination for the response.

Often, I've found myself manually implementing the extraction of `CorrelationData` and `ResponseTopic`, when these actions could likely be done by `Minimq` internally.

Perhaps we should adjust the API to make it more ergonomical to automatically forward-propagate these properties?

ryan-summers:
This PR takes out the dirty work of managing replying to an inbound message.

There is now a convenience `reply()` API that can be used. A new `ReplyOptions` can be constructed based on the received message properties and can be further used to customize `reply()` behavior, such as ignoring cases where there is no `ResponseTopic` or using a default `ResponseTopic` when none is found.

This PR fixes #84.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -2,6 +2,7 @@ use crate::{
     de::{received_packet::ReceivedPacket, PacketReader},
     network_manager::InterfaceHolder,
     packets::{ConnAck, Connect, PingReq, Pub, PubRel, SubAck, Subscribe},
+    reply_options::ReplyOptions,
     session_state::SessionState,
     types::{Properties, SubscriptionOptions, Utf8String},
     will::Will,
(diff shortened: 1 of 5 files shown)


src/mqtt_client.rs
```diff
@@ -385,6 +386,51 @@ impl<
...
+    /// * `retain` - The retain state of the response.
+    /// * `properties` - The properties to attach to the response.
+    pub fn reply<'a, const N: usize>(
+        &mut self,
+        options: ReplyOptions<'a>,
+        data: &[u8],
+        qos: QoS,
+        retain: Retain,
+        properties: &[Property<'a>],
+    ) -> Result<(), Error<TcpStack::Error>> {
```
jordens, line 408: Can we do this to automatically constrain `N`:

```suggestion
    pub fn reply<'a, const N: usize>(
        &mut self,
        options: ReplyOptions<'a>,
        data: &[u8],
        qos: QoS,
        retain: Retain,
        properties: &[Property<'a>; N],
    ) -> Result<(), Error<TcpStack::Error>> {
```

And then in the calling code either give a ref to the array or `client.reply(..., properties.try_into().unwrap())` using https://doc.rust-lang.org/std/primitive.array.html#method.try_from-1 ?

src/mqtt_client.rs
```diff
@@ -385,6 +386,51 @@ impl<
...
+    /// # Generics
+    /// * `N` specifies the maximum number of properties that the reply can contain.
+    ///
+    /// # Args
+    /// * [`ReplyOptions`] - Options used for constructing the reply.
+    /// * `default_response` - An optional default response topic to reply to.
+    /// * `data` - The data containing the response.
+    /// * `qos` - The QoS to transmit the response at.
+    /// * `retain` - The retain state of the response.
+    /// * `properties` - The properties to attach to the response.
```
jordens, line 400: Note that it will copy correlation data and that `N` needs to be (at least but preferrably exactly) `properties.len() + 1`. Or see below.

src/reply_options.rs
```diff
@@ -0,0 +1,61 @@
+use crate::{properties::Property, types::BinaryData};
+
+/// Used to specify information about replies to received MQTT messages.
+pub struct ReplyOptions<'a> {
```
jordens, line 4: Do we really need a struct for this? Can't this be just the original (or a new) slice of `&[Property]`?
That way you could pass the packet (or just the Property array/slice) to `reply()`.

jordens (changes requested):
If the main use-case of `reply` is to reply to an incoming message, can't we just pass that message's Properties to reply and let it do the right thing (determine N of initial plus additional properties, additional properties overriding the initial ones, and search for correlation data and response topic)?

I also wonder if we may want a builder interface to making packets (maybe just pub packets), instead of `client.publish()`

src/reply_options.rs
ryan-summers: I had it originally implemented like this, but found the desire to:
1. Add support for an optional-default topic if no `ResponseTopic` was found (e.g. default logging API)
2. Add an "ignore-message-if-no-`ResponseTopic`" functionality

Those would both just add arguments into the `reply()` API until it became unwieldy, so I opted to wrap them up in `ReplyOptions` instead

src/mqtt_client.rs
ryan-summers: Ah, then the vector could be `N + 1`, got it.

jordens: I imagine it would be `reply(..., properties.try_into::<&[Property; 4]>())`

ryan-summers: Why is that better than `reply<4>()`? The user still doesn't know how big `properties` is, but they now have more that they have to type

ryan-summers: Also, I believe that will then panic if you unwrap and `properties.len() != 4`

jordens: Agreed, it's not beautiful. But it replaces two checks and return paths for property length (L420, L426).

jordens (approved):
Let's land this and do the rest in #110

PR #108, opened 2022-09-05
Refactoring subscription API

### LINKED ISSUE #98: Add Subscription option flags support
Add support for subscription options, such as:
* NoLocal packet reception
* RetainAsPublished
* Retain handling

ryan-summers:
This PR fixes #98 by allowing for a list of `TopicFilter` objects to be provided to the `subscribe()` function. Each `TopicFilter` is composed of `SubscriptionOptions` and an `topic` path. Convenience `From<&str>` has been implemented for `TopicFilter` to allow the user to subscribe similar to the previous API.

@jordens There's some nuance in allowing the user to specify multiple TopicFilters per subscription request that ripples through the codebase. Let me know what you think about the change. We don't have to support multiple `TopicFilter`s per `SUBSCRIBE` if we don't want to.

### DIFF
--- a/src/de/received_packet.rs
+++ b/src/de/received_packet.rs
@@ -39,7 +39,10 @@ impl<'a> ReceivedPacket<'a> {
                 }
                 ReceivedPacket::SubAck(suback) => {
                     for code in remaining_payload.iter().map(|&x| ReasonCode::from(x)) {
-                        suback.codes.push(code).ok();
+                        suback
+                            .codes
+                            .push(code)
+                            .map_err(|_| ProtocolError::BufferSize)?;
                     }
                 }
                 _ => return Err(ProtocolError::MalformedPacket),
--- a/src/de/received_packet.rs
+++ b/src/de/received_packet.rs
@@ -1,8 +1,8 @@
 use crate::{
     message_types::MessageType,
     packets::{ConnAck, Disconnect, Pub, PubAck, PubComp, PubRec, SubAck},
-    varint::Varint,
     reason_codes::ReasonCode,
+    varint::Varint,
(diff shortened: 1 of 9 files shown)


src/de/received_packet.rs
```diff
@@ -213,7 +219,8 @@ mod test {
         let packet = ReceivedPacket::from_buffer(&serialized_suback).unwrap();
         match packet {
             ReceivedPacket::SubAck(sub_ack) => {
-                assert_eq!(sub_ack.code, ReasonCode::GrantedQos2);
+                assert_eq!(sub_ack.codes.len(), 1);
```
jordens, line 222: I think in general we should never panic on invalid messages. Just log/signal an error and continue.
If we can't skip through them without loosing packet sync, we should close the connection.
Panicing sounds decidedly wrong. But this is a general comment about all the code, not just this location here.

src/design_parameters.rs
```diff
@@ -0,0 +1,8 @@
+//! # Design Parameters
```
jordens, line 1: Maybe these can go right to the top into `crate`.

jordens (approved):
LGTM

src/de/received_packet.rs
ryan-summers: This is specifically a unit test, so we want things to panic on failure. However, I agree with you. I do not believe we panic internally in the library anywhere (or it would be a bug if we did).

It's an interesting point that maybe the software should handle all errors by restting the client and just telling the user that a `SessionReset` occurred.

jordens: Ah. I missed the context. All good!

PR #111, opened 2022-09-07
Adding MessageBuilder support

### LINKED ISSUE #110: Consider simplifying message construction
Consider simpler ways of constructing `Pub` messages for the user. It may make sense to allow the user to build a message locally that they provide to the `publish()` and `reply()` APIs directly, such as
```rust
// Construct a fresh message for publication.
let message = Publish(payload).topic("data/topic").qos(QoS::ExactlyOnce).properties(properties).finish().unwrap();

// Build a reply to a received packet.
let message = Publish(payload).reply(received_properties).finish().unwrap();
mqtt.client.publish(message);
```

ryan-summers:
This PR fixes #110 by refactoring how the `publish` function works and removes the `reply()` function.

Instead, the user can create a `Publication` object that builds a response or publication that can be provided to `publish()`.

`publish()` now takes a `Pub` packet type directly which is published via MQTT.
Example:
```rust
// Construct a new message
let publication = Publication::new(payload).topic("data/topic/tree").finish().unwrap();

// Construct a response to a received message, using a default topic if none is found.
let publication = Publication::new(payload).topic("my/default/topic").reply(received_properties).finish().unwrap();

// Construct a message with a specified QoS
let publication = Publication::new(payload).topic("data/topic/tree").qos(QoS::ExactlyOnce).finish().unwrap();

// Construct a retained message
let publication = Publication::new(payload).topic("data/topic/tree").retain().finish().unwrap();

// Publish the message.
mqtt.client().publish(publication).unwrap();
```

This PR builds on https://github.com/quartiq/minimq/pull/109


jordens:
I'll temporarily set the base branch to feature/subscription-qos to be able to review just the delta.

src/publication.rs
```diff
@@ -0,0 +1,135 @@
...
+
+        self
+    }
+
+    /// Generate the final publication.
+    ///
+    /// # Returns
+    /// The message to be published if a publication topic was specified. If no publication topic
+    /// was identified, an error is returned.
+    pub fn finish(self) -> Result<Pub<'a>, ProtocolError> {
```
jordens, line 124: Can't the builder pattern be implemented directly on `Pub` (get rid of `finish()` and the `Result::Err` path)?

jordens (approved):
Good.
The `CorrelatedProperties` variant starts to become a bit unwieldy though.

jordens:
Feel free to change the base branch back to master before merging.

src/publication.rs
ryan-summers: Ultimately, we have to handle the case when no `topic` is identified. This should be considered an anticipated case in my opinion, since someone who only expects to `reply()` to some other message would expect that the received properties would have a `ResponseTopic`. However, the received properties may not have this topic

jordens: Wouldn't both `Pub::new(topic) -> Self` and `Pub::reply(&self) -> Self` return structs with a valid topic?
I.e. move the payload setting (initialized as empty) into a builder method.

ryan-summers: No, they wouldn't. A `Pub::reply(&ReplyOptions)` still has the same problem where the `ResponseTopic` might not exist if the message you're replying to doesn't supply the property. What do you do when there's no `topic` to publish a message to? It's not something that can be sent over MQTT

src/types.rs
```diff
@@ -39,6 +45,23 @@ impl<'a> core::iter::Iterator for PropertiesIter<'a> {
...
+                    return Some(Ok(*correlation));
+                }
+
+                let index = self.index - 1;
+                if index >= properties.len() {
+                    return None;
+                }
+
+                Some(Ok(properties[index]))
+            }
```
ryan-summers, line 63: If we want to simplify iterating across properties, we can easily just only do an `IntoIter` implementation for `Properties::DataBlock` and raise `unimplemented!()` otherwise. I don't see any foreseeable case where a user would iterate across `Properties::Slice` or `Properties::CorrelatedSlice`

jordens (approved):

src/publication.rs
jordens: Well that would be `Pub::reply(&self) -> Option<Self>`, no?

PR #113, opened 2022-09-19
Removing arbitrary limit on subscription topic count

ryan-summers:
This PR removes the arbitrary restriction on the maximum number of topics per subscription by instead just taking a slice to the subscription reason codes. These codes are then processed individually in the handler instead of being converted into ReasonCodes and pushed into a slice.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -73,7 +73,11 @@ where
             Some(index) => self.session_state.pending_subscriptions.swap_remove(index),
         };
 
-        for code in subscribe_acknowledge.codes.iter() {
+        for code in subscribe_acknowledge
+            .codes
+            .iter()
+            .map(|&x| ReasonCode::from(x))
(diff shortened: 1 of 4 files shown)


src/mqtt_client.rs
```diff
@@ -73,7 +73,11 @@ where
...
         };
 
-        for code in subscribe_acknowledge.codes.iter() {
+        for code in subscribe_acknowledge
+            .codes
+            .iter()
+            .map(|&x| ReasonCode::from(x))
+        {
             code.as_result()?;
         }
```
jordens, line 82: Merging the `map` and loop body code:

```suggestion
        for &code in subscribe_acknowledge.codes.iter() {
            ReasonCode::from(x).as_result()?;
        }
```

jordens (approved):
Just a small simplification. Otherwise LGTM.

PR #123, opened 2022-12-14
Fixing user property serialization

### LINKED ISSUE #122: `UserProperty` is serialized backwards
The `UserProperty` is incorrectly serializing the `value` followed by the `key`, which means that the order is unintentionally swapped.

This is likely a holdover from when we used to invert serialization to simplify lengths.

ryan-summers:
This PR fixes #122 by correcting serialization of `UserProperty` properties. Previously, the key and value order were inverted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 This document describes the changes to Minimq between releases.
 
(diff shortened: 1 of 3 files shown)


jordens (approved):
Ack.

PR #126, opened 2023-05-11
Fixing MQTT permanent disconnection issues

### LINKED ISSUE #125: Multiple socket disconnections may cause the client to fail to reconnect
If the TCP socket connection is lost immediately after the socket is opened, but before the `Connect` packet can be sent, the state machine gets stuck in a way that means that the MQTT client will never reconnect to the broker.

The only method to force a reconnection would be a forced power cycle of the device.

Identified in https://github.com/quartiq/stabilizer/issues/674

ryan-summers:
This PR fixes #125 

TODO:
- [x] Update `std-embedded-nal` + integration tests
- [x] Wait for https://github.com/rust-embedded-community/embedded-nal/pull/85 to release

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -436,7 +436,7 @@ impl<
 
     fn update(&mut self) -> Result<(), Error<TcpStack::Error>> {
         if self.network.socket_was_closed() {
-            warn!("Handling closed socket");
+            info!("Handling closed socket");
             self.sm.process_event(Events::TcpDisconnect).unwrap();
             self.network.allocate_socket()?;
         }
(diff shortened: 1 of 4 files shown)


src/mqtt_client.rs
```diff
@@ -438,21 +435,21 @@ impl<
     }
 
     fn update(&mut self) -> Result<(), Error<TcpStack::Error>> {
-        // Potentially update the state machine depending on the current socket connection status.
-        let tcp_connected = self.network.tcp_connected()?;
-
-        if !tcp_connected {
-            self.sm.process_event(Events::TcpDisconnect).ok();
+        if self.network.socket_was_closed() {
+            warn!("Handling closed socket");
```
jordens, line 439: I'd leave this at info. Disconnects are normal operation.

jordens (approved):

PR #129, opened 2023-07-21
Adding QoS::ExactlyOnce RX support

### LINKED ISSUE #62: Allow subscribing at `QoS::ExactlyOnce`
Currently, QoS support is only implemented for sending QoS > 0 messages, but not for receiving. We need to respond with PubAck if the server ever sends us a publish message with Qos > 0 as well

ryan-summers:
Fixes #62 by allowing subscriptions at QoS::ExactlyOnce

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -558,41 +558,44 @@ impl<
                 match info.qos {
                     QoS::AtMostOnce => {}
                     QoS::AtLeastOnce => {
-                        let puback = PubAck {
-                            // Note(uwnrap): There should always be a packet ID for QoS >
-                            // AtMostOnce.
-                            packet_identifier: info.packet_id.unwrap(),
+                        // Note(uwnrap): There should always be a packet ID for QoS > AtMostOnce.
+                        let packet_id = info.packet_id.unwrap();
 
-                            // Note: Because we do not support ExactlyOnce, it's not possible for
-                            // us to receive two packets with the same ID.
-                            reason: ReasonCode::Success.into(),
+                        // Reject the packet ID if it's currently in use for another publication.
+                        let reason = if self
+                            .sm
+                            .context_mut()
+                            .session_state
+                            .pending_packets
+                            .iter()
+                            .any(|&id| id == packet_id)
+                        {
+                            ReasonCode::PacketIdInUse
+                        } else {
+                            ReasonCode::Success
+                        };
+
+                        let puback = PubAck {
+                            packet_identifier: info.packet_id.unwrap(),
+                            reason: reason.into(),
                         };
 
                         self.network.send_packet(&puback)?;
                     }
 
                     QoS::ExactlyOnce => {
+                        let session_state = &mut self.sm.context_mut().session_state;
                         // Note(uwnrap): There should always be a packet ID for QoS >
                         // AtMostOnce.
                         let packet_id = info.packet_id.unwrap();
 
                         // Check if the packet ID already exists before forwarding to app
-                        let duplicate = self
-                            .sm
-                            .context_mut()
-                            .session_state
+                        let duplicate = session_state
                             .pending_packets
                             .iter()
                             .any(|&x| x == packet_id);
 
-                        let reason = if self
-                            .sm
-                            .context_mut()
-                            .session_state
-                            .pending_packets
-                            .push(packet_id)
-                            .is_err()
-                        {
+                        let reason = if session_state.pending_packets.push(packet_id).is_err() {
                             ReasonCode::ReceiveMaxExceeded
                         } else {
                             ReasonCode::Success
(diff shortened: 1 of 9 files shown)


src/session_state.rs
```diff
@@ -31,6 +31,7 @@ pub struct SessionState<
     pub maximum_packet_size: Option<u32>,
     pub client_id: String<64>,
     pub pending_subscriptions: Vec<u16, 32>,
+    pub pending_packets: Vec<u16, MSG_COUNT>,
```
ryan-summers, line 34: This should be combined with `pending_messages` etc. in a single buffer.

src/packets.rs
```diff
@@ -202,7 +200,7 @@ impl<'a> From<ReasonCode> for Reason<'a> {
         Self {
             reason: Some(ReasonData {
                 code,
-                _properties: None,
+                _properties: Some(Properties::Slice(&[])),
```
jordens, line 203: Out of curiosity: Why that change? ~Can't we serialize `None`?~

src/session_state.rs
jordens: was about to write that.

src/mqtt_client.rs
```diff
@@ -536,8 +571,47 @@ impl<
                         self.network.send_packet(&puback)?;
                     }
 
-                    // TODO: Add support.
-                    QoS::ExactlyOnce => unimplemented!(),
+                    QoS::ExactlyOnce => {
+                        // Note(uwnrap): There should always be a packet ID for QoS >
+                        // AtMostOnce.
+                        let packet_id = info.packet_id.unwrap();
```
jordens, line 577: Doesn't this come from the broker?
We should never panic on erroneous broker data. Either handle, ignore, or reset.

src/mqtt_client.rs
```diff
@@ -514,6 +522,33 @@ impl<
...
+                    ReasonCode::Success
+                } else {
+                    ReasonCode::PacketIdNotFound
+                };
+
+                // Send a PubComp
+                let pubcomp = PubComp {
+                    packet_id: rel.packet_id,
+
+                    // Note: Because we do not support ExactlyOnce, it's not possible for
```
jordens, line 544: I thought this implements ExactlyOnce.

src/mqtt_client.rs
```diff
@@ -536,8 +571,47 @@ impl<
...
                     }
 
-                    // TODO: Add support.
-                    QoS::ExactlyOnce => unimplemented!(),
+                    QoS::ExactlyOnce => {
+                        // Note(uwnrap): There should always be a packet ID for QoS >
+                        // AtMostOnce.
+                        let packet_id = info.packet_id.unwrap();
+
+                        // Check if the packet ID already exists before forwarding to app
```
jordens, line 579: Is it certain that the broker will PubRel also the duplicate packet_ids in pending_packets?

src/packets.rs
ryan-summers: There's some strangeness in `None` for the serializer because MQTT allows for things to be entirely omitted. The serializer treats `None` as _literally nothing to serialize_. So no data gets serialized at all. However, the MQTT spec requires that we serialize the varint of the properties length (a single zero). So in this case, `None` serializes to `[]`, but `Some(Properties::Slice(&[]))` serializes to `[0]`. It's funky, but flexible.

src/mqtt_client.rs
ryan-summers: This must be a copy-paste fail. I'll take a look at this and figure out what the comment is referring to

src/mqtt_client.rs
ryan-summers: This doesn't come from the broker, it comes from our deserializer. The QoS bits of the control byte tell us whether or not the ID is present in the message. So if QoS > AtMostOnce, there's always a packet_id in the option.

I agree we should never panic based on external input

src/mqtt_client.rs
ryan-summers: The broker is required to PubRel the packet_id after receiving the PubRec in order to release the ID for further use. If the broker were to reuse the ID without a preceding PubRel, it would be a protocol violation.

jordens: Then I don't get the intention behind the handling of duplicate packet_ids in pending_packets here.

ryan-summers: The broker may have disconnected before it received our PubRec and we're maintaining the session state across a reconnection. This is taken directly from MQTT 4.3.3 and 4.4

src/session_state.rs
ryan-summers: Actually, MQTT specifies that the client and server have independent sets of packet IDs. Given that this represents the server-provided packet-ids, it is important we separate these. I'll make this more clear with naming etc.

src/mqtt_client.rs
jordens: Ack. Might be good idea to ref the spec in the code where possible.

src/session_state.rs
jordens: Ok. Then `pending_messages()` in `SessionState` and `MqttClient` is ambiguous as it conflates them.
And it may be wise to be very specific and rename the vec to `pending_pubrel`.

ryan-summers:
@jordens This should be good to go

PR #130, opened 2023-07-21
Updating client to respect server receive maximum

ryan-summers:
This PR updates the client to respect the server receive maximums.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -48,6 +48,7 @@ struct ClientContext<
     const MSG_COUNT: usize,
 > {
     session_state: SessionState<TcpStack, Clock, MSG_SIZE, MSG_COUNT>,
+    send_quota: u16,
 }
 
 impl<TcpStack, Clock, const MSG_SIZE: usize, const MSG_COUNT: usize>
@@ -108,6 +109,7 @@ where
             ReceivedPacket::PubAck(ack) => {
                 // No matter the status code the message is considered acknowledged at this point
                 self.session_state.handle_puback(ack.packet_identifier)?;
+                self.send_quota.saturating_add(1);
             }
             _ => return Err(Error::Protocol(ProtocolError::UnsupportedPacket)),
         }
@@ -127,6 +129,9 @@ where
             self.session_state.reset();
         }
 
+        // If the server doesn't specify a send quota, assume it's 65535 as required by the spec.
+        self.send_quota = u16::MAX;
+
         for property in acknowledge.properties.into_iter() {
             match property? {
                 Property::MaximumPacketSize(size) => {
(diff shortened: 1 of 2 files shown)


src/mqtt_client.rs
```diff
@@ -379,10 +397,9 @@ impl<
 
         // TODO: Generate event.
         if let Some(id) = publish.packet_id {
-            self.sm
-                .context_mut()
-                .session_state
-                .handle_publish(publish.qos, id, packet)?;
+            let context = self.sm.context_mut();
+            context.session_state.handle_publish(publish.qos, id, packet)?;
+            context.send_quota = context.send_quota.saturating_sub(1);
```
ryan-summers, line 402: Not sure if this should be saturating or not. Technically, it would be a defect in minimq if we tried to decrease beyond zero

jordens: `checked_sub().unwrap()` then?

src/mqtt_client.rs
```diff
@@ -127,6 +129,11 @@ where
             self.session_state.reset();
         }
 
+        // If the server doesn't specify a send quota, assume it's 65535 as required by the spec
+        // section 3.1.2.11.3 - this value is not part of the session state and is reset for each
```
jordens, line 133: Isn't the section 3.2.2.3.3?

src/mqtt_client.rs
```diff
@@ -108,6 +109,7 @@ where
             ReceivedPacket::PubAck(ack) => {
                 // No matter the status code the message is considered acknowledged at this point
                 self.session_state.handle_puback(ack.packet_identifier)?;
+                self.send_quota.saturating_add(1);
```
jordens, line 112: As per [4.9](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901251), don't you also need to replenish on PubComp and PubRec-Error?

ryan-summers: Indeed - I got distracted half way through implementation - thanks for pointing this out. Should be cleaned up now.

src/mqtt_client.rs
```diff
@@ -104,9 +114,15 @@ where
         match &packet {
             ReceivedPacket::SubAck(ack) => self.handle_suback(ack)?,
             ReceivedPacket::PingResp => self.session_state.register_ping_response(),
-            ReceivedPacket::PubComp(comp) => self.session_state.handle_pubcomp(comp.packet_id)?,
+            ReceivedPacket::PubComp(comp) => {
+                self.send_quota =
+                    core::cmp::min(self.send_quota.saturating_add(1), self.max_send_quota);
```
jordens, line 119:
```suggestion
                    self.send_quota.saturating_add(1).min(self.max_send_quota);
```

and in the two other places as well.

jordens (changes requested):

src/mqtt_client.rs
ryan-summers: I was trying to find this in docs and couldn't for some reason. This is exactly what I want, thanks

jordens (approved):

PR #131, opened 2023-07-21
Using user-supplied RX and TX buffers, removing const-generics

### LINKED ISSUE #92: Refactor session state message storage
Currently, messages with QoS > AtMostOnce are stored in the `SessionState`. MQTT requires that we republish these messages upon reconnection with the broker in the same order as originally published.

Currently, we store the messages in two datastructures. One is a vector of message IDs (in order) and the other is a map of the messages (where the ID is the key to the map).

It would be desirable to combine these into a single datastructure. Some requirements:
* Removing an element must not result in large copies, as this may result in copying far too much memory if there's a lot of pending messages
* A single data structure should be used
* An ideal implementation may allow us to not hold any additional buffer space (e.g. when QoS AtMostOnce is used exclusively).

### LINKED ISSUE #112: Const-generics can cause flash requirement bloat
Because of the usage of const-generics on the `MqttClient` and `SessionState`, there's a number of code that gets generated twice in final binaries. This can result in significant binary bloat when using multiple clients with e.g. different maximum message sizes:

Here's an abbreviated `cargo nm --bin dual-iir --release -- -S --size-sort | grep minimq` output for `stabilizer

| Size (Hex) | Type | Name | Size (dec) |
-- | -- | -- | --
00000190 | t | minimq::session_state::SessionState<TcpStack,Clock,_,_>::remove_packet::h9eb457d89b5ba34f | 400
00000196 | t | minimq::session_state::SessionState<TcpStack,Clock,_,_>::remove_packet::hb86daa97efcb21a3 | 406
000002da | t | minimq::network_manager::InterfaceHolder<TcpStack,_>::write::hf41deed262d53273 | 730
000002e8 | t | minimq::network_manager::InterfaceHolder<TcpStack,_>::write::h754f7308e9ad13ad | 744
000005b6 | t | minimq::mqtt_client::MqttClient<TcpStack,Clock,_,_>::publish::h6f4cf5e5c2f6e63f | 1462
000006b4 | t | minimq::de::received_packet::ReceivedPacket::from_buffer::hca71e2b28605d17d | 1716
000006b4 | t | <minimq::types::PropertiesIter as   core::iter::traits::iterator::Iterator>::next::h7e3a709dabea9b6c | 1716
0000070c | t | minimq::mqtt_client::sm::StateMachine<T>::process_event::hb877959fab492fa2 | 1804
0000076e | t | minimq::mqtt_client::sm::StateMachine<T>::process_event::hbf9c983f23ba6d6c | 1902

As can be seen in the above, there's two definitions for `StateMachine<T>::process_event`, `InterfaceHolder::write`, and `SessionState::remove_packet` that cause non-trivial flash consumption.

ryan-summers:
This is a first stab at updating minimq to remove const generics and take the smoltcp path of user-supplied buffers. This should drastically cut down on stack and RAM usage for end users.

Fixes #112

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -89,22 +89,29 @@ impl<'a> MqttSerializer<'a> {
     /// # Args
     /// * `buf` - The buffer to encode data into.
     /// * `packet` - The packet to encode.
+    pub fn to_buffer_meta<T: Serialize + ControlPacket>(
+        buf: &'a mut [u8],
+        packet: &T,
+    ) -> Result<(usize, &'a [u8]), Error> {
+        let mut serializer = Self::new(buf);
+        packet.serialize(&mut serializer)?;
+        let (offset, packet) = serializer.finalize(T::MESSAGE_TYPE, packet.fixed_header_flags())?;
+        Ok((offset, packet))
+    }
+
+    /// Encode an MQTT control packet into a buffer.
+    ///
+    /// # Args
+    /// * `buf` - The buffer to encode data into.
+    /// * `packet` - The packet to encode.
     pub fn to_buffer<T: Serialize + ControlPacket>(
         buf: &'a mut [u8],
         packet: &T,
     ) -> Result<&'a [u8], Error> {
         let mut serializer = Self::new(buf);
         packet.serialize(&mut serializer)?;
-        let packet = serializer.finalize(T::MESSAGE_TYPE, packet.fixed_header_flags())?;
+        let (_, packet) = serializer.finalize(T::MESSAGE_TYPE, packet.fixed_header_flags())?;
         Ok(packet)
-    }
-
-    /// Finish the packet without prepending the MQTT fixed header.
-    ///
-    /// # Returns
-    /// A slice representing the serialized packet without a fixed header.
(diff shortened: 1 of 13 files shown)


ryan-summers:
@jordens I'm still working on removing the const generics from the session state, but I think that's going to be a larger follow-up PR, so let's land this one as-is.

src/ser/mod.rs
```diff
@@ -89,22 +89,29 @@ impl<'a> MqttSerializer<'a> {
...
+    /// * `buf` - The buffer to encode data into.
+    /// * `packet` - The packet to encode.
+    pub fn to_buffer<T: Serialize + ControlPacket>(
+        buf: &'a mut [u8],
+        packet: &T,
+    ) -> Result<&'a [u8], Error> {
+        let mut serializer = Self::new(buf);
+        packet.serialize(&mut serializer)?;
+        let (_, packet) = serializer.finalize(T::MESSAGE_TYPE, packet.fixed_header_flags())?;
+        Ok(packet)
```
jordens, line 114: can't you call into `to_buffer_meta()` here?

jordens (commented):
Looks nice!

jordens (approved):

PR #139, opened 2023-08-02
Refactoring session state buffer management

### LINKED ISSUE #92: Refactor session state message storage

ryan-summers:
This PR fixes #92 by refactoring the session state to use a singular ring buffer supplied by the user for maintaining republication state.

This should reduce code size and RAM memory footprint.

Now, the session state is backed primarily with a user-provided `session` buffer. This buffer is used for storing `publish` messages with QoS > 0 until they are acked.

### DIFF
--- a/src/de/received_packet.rs
+++ b/src/de/received_packet.rs
@@ -11,32 +11,6 @@ use bit_field::BitField;
 use core::convert::TryFrom;
 use serde::Deserialize;
 
-pub struct Packet<'a> {
-    length: Varint,
-
-    // Note: If deserialized from a split buffer, the contents of this packet may not be correct.
-    // As such, we do not expose it in the public API.
-    packet: ReceivedPacket<'a>,
-}
-
-impl<'a> Packet<'a> {
-    pub fn len(&self) -> usize {
-        self.length.len() + 1 + self.length.0 as usize
-    }
-
-    pub fn id(&self) -> Option<u16> {
-        match &self.packet {
-            ReceivedPacket::Publish(info) => info.packet_id,
-            ReceivedPacket::PubAck(ack) => Some(ack.packet_identifier),
-            ReceivedPacket::SubAck(ack) => Some(ack.packet_identifier),
-            ReceivedPacket::PubRel(rel) => Some(rel.packet_id),
-            ReceivedPacket::PubRec(rec) => Some(rec.packet_id),
-            ReceivedPacket::PubComp(comp) => Some(comp.packet_id),
-            _ => None,
-        }
-    }
-}
-
 #[derive(Debug)]
 pub enum ReceivedPacket<'a> {
     ConnAck(ConnAck<'a>),
@@ -53,14 +27,13 @@ pub enum ReceivedPacket<'a> {
 impl<'a> ReceivedPacket<'a> {
     pub fn from_buffer(buf: &'a [u8]) -> Result<Self, ProtocolError> {
         let mut deserializer = MqttDeserializer::new(buf);
-        let mut packet = Packet::deserialize(&mut deserializer)?;
+        let mut packet = ReceivedPacket::deserialize(&mut deserializer)?;
 
-        // Remainder should never error because there is no tail data.
-        let remaining_payload = deserializer.remainder().unwrap();
+        let remaining_payload = deserializer.remainder();
 
         // We should only have remaining payload for publish messages.
         if !remaining_payload.is_empty() {
-            match &mut packet.packet {
+            match &mut packet {
                 ReceivedPacket::Publish(publish) => {
                     publish.payload = remaining_payload;
                 }
@@ -71,20 +44,6 @@ impl<'a> ReceivedPacket<'a> {
             }
         }
 
-        Ok(packet.packet)
-    }
-
-    /// Get a received packet from a split buffer
-    ///
-    /// # Note
-    /// Packets deserialized in this manner may not have valid binary slices. This occurs because
-    /// the binary slice may span the discontinuity, and thus it's impossible to zero-copy it.
-    /// Instead, the resulting binary slice will only contain the maximum amount of continuous
-    /// data.
-    pub fn from_split_buffer(buf: &'a [u8], tail: &'a [u8]) -> Result<Packet<'a>, ProtocolError> {
-        let mut deserializer = MqttDeserializer::new_split(buf, tail);
-        deserializer.truncate_on_discontinuity();
-        let packet = Packet::deserialize(&mut deserializer)?;
         Ok(packet)
     }
 }
@@ -92,7 +51,7 @@ impl<'a> ReceivedPacket<'a> {
 struct ControlPacketVisitor;
 
 impl<'de> serde::de::Visitor<'de> for ControlPacketVisitor {
-    type Value = Packet<'de>;
+    type Value = ReceivedPacket<'de>;
 
     fn expecting(&self, formatter: &mut core::fmt::Formatter) -> core::fmt::Result {
         write!(formatter, "MQTT Control Packet")
@@ -105,7 +64,7 @@ impl<'de> serde::de::Visitor<'de> for ControlPacketVisitor {
         // always providing us some new element or an error that we return based on our
         // deserialization implementation.
         let fixed_header: u8 = seq.next_element()?.unwrap();
-        let length: Varint = seq.next_element()?.unwrap();
+        let _length: Varint = seq.next_element()?.unwrap();
         let packet_type = MessageType::try_from(fixed_header.get_bits(4..=7))
             .map_err(|_| A::Error::custom("Invalid MQTT control packet type"))?;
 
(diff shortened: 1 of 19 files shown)


ryan-summers:
@jordens This is ready for an initial review, but I'm going to add some more testing around the ring buffer and republication semantics. I'd love to get your opinion on the implementation though.

src/republication.rs
```diff
@@ -0,0 +1,164 @@
...
+            if index > 1 {
+                self.pubrel_republish_index.replace(index - 1);
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn push_pubrel(&mut self, pubrel: &PubRel) -> Result<(), ProtocolError> {
+        self.pending_pubrel
```
ryan-summers, line 102: The `pubrel` list is now a fixed size, but `publish` is variable in length. This means that the `pubrel` list could fill up if we send out too many `publish` messages to the broker. Perhaps we should encode the pending `pubrel` messages directly into the ring buffer as well for republication..

However, I don't think MQTT guarantees ordering between the PUBLISH and the PUBREL messages, so we might get a PUBCOMP message before we get a PUBREC if we send a PUBLISH followed by a PUBREL to the server. Because the ring buffer requires ordering, this would break things.... Need to think about this further

ryan-summers: That's a really good point, thanks for pointing that out. Simplifies the design.

src/network_manager.rs
```diff
@@ -133,11 +138,12 @@ where
...
 
         crate::info!("Sending: {:?}", packet);
-        let (offset, packet) = crate::ser::MqttSerializer::to_buffer_meta(self.tx_buffer, packet)?;
+        let (offset, packet) = crate::ser::MqttSerializer::to_buffer_meta(self.tx_buffer, packet)
+            .map_err(ProtocolError::from)?;
         let len = packet.len();
 
         self.commit_write(offset, len)?;
-        Ok(&self.tx_buffer[offset..len])
+        Ok(&self.tx_buffer[offset..][..len])
```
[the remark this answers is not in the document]
ryan-summers: Yeah it appears so. I'm unsure why we didn't hit it previously

src/republication.rs
```diff
@@ -0,0 +1,164 @@
...
+            return Err(ProtocolError::UnexpectedPacket);
+        };
+
+        if *pending_id != id {
+            return Err(ProtocolError::BadIdentifier);
+        }
+
+        // Now that we received the PubComp for this PubRel, we can remove it from our session
+        // state. We will not need to retransmit this upon reconnection.
+        self.pending_pubrel.remove(0);
```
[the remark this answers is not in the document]
ryan-summers: The problem is that the ring buffer needs borrowed memory, which we can't hand to Deque because heapless uses const-generics and owned memory instead.

I've updated the pending_pubrel to use a deque though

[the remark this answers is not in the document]
jordens: Still maybe make the RingBuffer API mirror the Deque API as much as possible. That forces thinking about the same corner cases and resolving the same questions.

src/ring_buffer.rs
```diff
@@ -0,0 +1,116 @@
+use crate::de::received_packet::ReceivedPacket;
+
+pub(crate) struct RingBuffer<'a> {
```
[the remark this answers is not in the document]
ryan-summers: It was a TODO for me. Didn't have the time to write any beforehand. I just added some nice unit tests for corner cases that I could think of.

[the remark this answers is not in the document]
ryan-summers: Also, there were many bugs

src/ring_buffer.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    pub fn pop(&mut self, len: usize) {
+        let len = len.max(self.count);
+
+        self.head = (self.head + len).rem_euclid(self.data.len());
+        self.count -= len;
+    }
+
+    pub fn probe_header(&mut self, index: usize) -> Option<(usize, MqttHeader)> {
```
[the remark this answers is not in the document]
ryan-summers: I removed the MQTT-specific stuff from the RingBuffer implementation

src/republication.rs
```diff
@@ -49,11 +54,40 @@ impl<'a> RepublicationBuffer<'a> {
         }
     }
 
-    pub fn pop_publish(&mut self, id: u16) -> Result<(), ProtocolError> {
-        let (_, header) = self
+    fn probe_header(&mut self, index: usize) -> Option<(usize, MqttHeader)> {
+        let mut offset = 0;
+
+        if index > 0 {
```
jordens, line 60: The `if` can go and this can become one loop `for _ in ..index + 1 { ... break packet; }` or also move the `Some((offset..))` into the loop as well.

src/network_manager.rs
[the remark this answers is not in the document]
jordens: If offset is always 0 in the unittests, then it wouldn't hit.

src/republication.rs
ryan-summers: You can't break out of for-loops, but I've simplified nonetheless

src/de/received_packet.rs
```diff
@@ -44,14 +68,20 @@ impl<'a> ReceivedPacket<'a> {
             }
         }
 
+        Ok(packet.packet)
+    }
+
+    pub fn from_split_buffer(buf: &'a [u8], tail: &'a [u8]) -> Result<Packet<'a>, ProtocolError> {
+        let mut deserializer = MqttDeserializer::new_split(buf, tail);
```
ryan-summers, line 75: I think this breaks if there's a data chunk spanning the head/tail divide. Perhaps we only want to deserialize the header of the packet, which can't break.

ryan-summers:
This should now have all the kinks worked out. I gave it a spin on stabilizer just to be sure, although that isn't using any of the QoS stuff

src/de/received_packet.rs
```diff
@@ -44,14 +71,28 @@ impl<'a> ReceivedPacket<'a> {
...
+        Ok(packet.packet)
+    }
+
+    /// Get a received packet from a split buffer
+    ///
+    /// # Note
+    /// Packets deserialized in this manner may not have valid binary slices. This occurs because
+    /// the binary slice may span the discontinuity, and thus it's impossible to zero-copy it.
+    /// Instead, the resulting binary slice will only contain the maximum amount of continuous
+    /// data.
```
jordens, line 83: Being worried about how one would then go about actually getting the complete valid packet, it took me a while to figure out that this path is only used in republication, where we don't really need to deserialize anything other than the header for probing.
Wouldn't `header_only()` (and then returning after the header, or even only returning the header) instead of `truncate_on_discontinuity()` be more adequate?
And what happens if the header is split across the buffers? Does it still deserialize that correctly (since no borrowed data)?

ryan-summers: There's a few things to note:
1. When deserializing via the discontinuous buffer, the complete `packet` deserialized is _not_ `pub` to the crate, so it's not possible to unintentionally use invalid contents. The API only exposes `id()` and `len()` functions for these packets.
2. `header_only()` would be ideal, but unfortunately we need to get the `Pub` packet ID, which _may or may not_ be present for packets. So we have to do a complete fixed header deserialization, check the packet type, and then deserialize the `Pub` packet. We can go this route, but it's a lot of duplicated code
3. It's not relevant if the header is split across the slices because the header information contains only primitive types. That being said, I'm now seeing an issue where even u16s have to be within the continuous memory, will fix

jordens: Ok. That kills the first alternative. What about the other (`Deque<(offset, MqttHeader), 10>` as an index into the `RingBuffer`)? It would only be slightly more code for managing that and a bit more stack but would remove the repeated deserialization and allow `take(n)` again in the deseroializer. It wouldn't duplicate any deser code.

ryan-summers: I got it even simpler with a single `Deque<offset, 10>` - no need for any deserialization until we're about to pop the packet.

ryan-summers:
@jordens Ping on this - it should be ready to go now

src/de/received_packet.rs
jordens: Nice. But there is still the one unnecessary deserialization in `pop_publish`. This is a packet that we serialized when it was added to the ring buffer, i.e. we know its packet length and packet id. Wouldn't removing that deserialization step (in favor of maintaining a `Deque<(usize, MqttHeader)>`) remove all the "deserialize from split buffer" gymnastics in `deserializer.rs`?

jordens (commented):
i still wonder whether the "deserialize from split buffer" logic is all necessary (see https://github.com/quartiq/minimq/pull/139#discussion_r1296872170). Keeping track of `Deque<(length, packet_id)>` for the republication ringbuffer should allow getting rid of it, no?

ryan-summers:
Yeah that technically works just fine. There's a little less protection against code defects though because we never check the republication buffer, so it's possible that a defect that corrupts the buffer may mess with our republication state.

However, if that happened, we'd cause protocol errors on the TCP socket and everything would reset anyways, so it's likely not a big deal. It's not updated.

src/republication.rs
```diff
@@ -128,6 +110,24 @@ impl<'a> RepublicationBuffer<'a> {
...
     }
 
+    fn probe_header(&mut self, index: usize) -> Option<(usize, MqttHeader)> {
+        let offset = self
+            .pending_pub
+            .iter()
+            .take(index)
+            .map(|(_, len)| len)
+            .sum();
+        let (id, len) = self.pending_pub.iter().nth(index).unwrap();
```
jordens, line 120: Interesting! I tried to put this into a single iter with `fold()` but it's much less comprehensible.

jordens (approved):
Nice!

PR #140, opened 2023-08-21
Adding support for authentication

### LINKED ISSUE #119: Add authentication support
Add support for providing authentication credentials as part of the connection process

cnmozzie:
Add a set_auth function to set the user name and password. I test with the version 0.7.0 and it works fine. Not sure if it compatible with the current version.
[issue 119](https://github.com/quartiq/minimq/issues/119)

Fixes #119

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -208,6 +208,7 @@ impl<
     /// # Args
     /// * `user_name` - The user name
     /// * `password` - The password
+    #[cfg(feature = "unsecure")]
     pub fn set_auth(
         &mut self,
         user_name: &str,
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -4,7 +4,7 @@ use crate::{
     packets::{ConnAck, Connect, PingReq, Pub, PubAck, PubRel, SubAck, Subscribe},
     reason_codes::ReasonCode,
     session_state::SessionState,
-    types::{Properties, TopicFilter, Utf8String},
+    types::{Properties, TopicFilter, Utf8String, Auth},
     will::Will,
     Error, Property, ProtocolError, QoS, Retain, MQTT_INSECURE_DEFAULT_PORT,
     {debug, error, info, warn},
@@ -168,6 +168,7 @@ pub struct MqttClient<
     sm: sm::StateMachine<ClientContext<TcpStack, Clock, MSG_SIZE, MSG_COUNT>>,
     network: InterfaceHolder<TcpStack, MSG_SIZE>,
     will: Option<Will<MSG_SIZE>>,
+    auth: Option<Auth>,
     broker: SocketAddr,
 }
 
@@ -199,6 +200,24 @@ impl<
         will.qos(qos);
 
         self.will.replace(will);
+        Ok(())
+    }
+
+    /// Specify the authentication message used by the server.
+    ///
+    /// # Args
+    /// * `user_name` - The user name
+    /// * `password` - The password
+    pub fn set_auth(
+        &mut self,
+        user_name: &str,
+        password: &str,
+    ) -> Result<(), Error<TcpStack::Error>> {
+        let auth = Auth { 
+            user_name: String::from_str(user_name).or(Err(Error::ProvidedClientIdTooLong))?, 
+            password: String::from_str(password).or(Err(Error::ProvidedClientIdTooLong))?, 
+        };
+        self.auth = Some(auth);
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


src/mqtt_client.rs
```diff
@@ -202,6 +203,24 @@ impl<
         Ok(())
     }
 
+    /// Specify the authentication message used by the server.
+    ///
+    /// # Args
+    /// * `user_name` - The user name
+    /// * `password` - The password
+    pub fn set_auth(
```
ryan-summers, line 211: I'm hesitant about this because it lets a person provide credentials without enforcing the usage of TLS. These parameters should probably be only accepted IFF we are on the TLS port of MQTT. It would be a big security vulnerability to allow people to transmit plaintext auth credentials over the unsecure MQTT port

src/types.rs
```diff
@@ -167,6 +169,12 @@ impl<'de> serde::de::Deserialize<'de> for BinaryData<'de> {
     }
 }
 
+#[derive(Debug)]
+pub struct Auth {
+    pub user_name: String<64>,
+    pub password: String<64>,
```
ryan-summers, line 175: Do we have to have these as stored strings? Can we instead just borrow the memory from the user? The problem with this is that it's imposing a 128 byte overhead on all users of Minimq regardless of whether or not they're using auth. Not sure if this matters, but it would be nice to keep our memory overhead to a minimum.

src/mqtt_client.rs
ryan-summers: Maybe we can move these parameters as an option for setting the MQTT secured port?

ryan-summers (commented):
Thanks for the changes! Auth is definitely something we want to add - I haven't looked into it because I don't have an embedded-tls stack operational yet.

src/types.rs
cnmozzie: I agree. Because I am new in rust I refer to the code you wrote elsewhere:
[src/session_state.rs](https://github.com/quartiq/minimq/blob/8e7e2b8c6faf72b87a630b0c2a0571a8d117f187/src/session_state.rs#L11)

src/mqtt_client.rs
cnmozzie: I understand your concern. However, a lot of clients allow people to provide credentials without enforcing the usage of TLS, like [paho-mqtt](https://pypi.org/project/paho-mqtt/). And sometimes people have to connect to a broker that need  credentials. I think maybe we can turn it as a feature and disabled by default. For users who need this function, they can enable this feature anyway.

src/types.rs
ryan-summers: You can just take a `&'a str` instead, although this may require you to merge in `main` branch. Then, when constructing the CONNECT packet, you just wrap the &str in the `Utf8String`. If you want to leave this as-is, I can patch it up after the merge as well and tag you for review :)

src/mqtt_client.rs
ryan-summers: That's.... fair. I just don't think it's a great idea to give engineers a loaded gun pointed directly at their foot. Sure, some may knowingly be doing this because they don't care about keeping the credentials secret, but others may be oblivious to the massive cybersecurity vulnerability that this exposes.

ryan-summers: In any case, we should have the ability to set authentication credentials on an unsecure channel **disabled by default**. I'm fine with allowing people to opt-in if they acknowledge the risk. There _are_ real world use cases where this makes sense, i.e. if you have a private LAN.

ryan-summers:
@cnmozzie Are you still interested in this functionality? I was planning on cutting a release of minimq soon and it would be nice to have this in there :)

src/mqtt_client.rs
cnmozzie: I just add the "unsecure" feature to the "set_auth" function so that it is disabled by default

src/types.rs
cnmozzie: I prefer you to patch it up after the merge. I think you can handle it better than me.

ryan-summers: Actually, lets leave these as static strings for now. I wonder if the `'buf` lifetime will cause problems for users

ryan-summers (approved):
Looking good! Thanks! :)

src/mqtt_client.rs
```diff
@@ -277,6 +278,25 @@ pub struct MqttClient<
 impl<'buf, TcpStack: TcpClientStack, Clock: embedded_time::Clock, Broker: crate::Broker>
     MqttClient<'buf, TcpStack, Clock, Broker>
 {
+    /// Specify the authentication message used by the server.
+    ///
+    /// # Args
+    /// * `user_name` - The user name
+    /// * `password` - The password
+    #[cfg(feature = "unsecure")]
+    pub fn set_auth(
```
ryan-summers, line 287: This should go on the new `ConfigBuilder` instead now

ryan-summers: I'll handle this refactor in a follow-on PR :)

ryan-summers:
I think I did a mismerge through the Github UI on accident. I'm going to emrge as-is and manually fix it in a follow-up

PR #141, opened 2023-08-22
Adding support for server maximum QoS

### LINKED ISSUE #136: Respect server "Maximum QoS"
"The server may indicate a maximum QoS that it supports. We should respect this value and never transmit at an excessive QoS

ryan-summers:
Fixes #136 by allowing auto-downgrade of publications to the max QoS supported by the server if desired.

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -10,7 +10,7 @@ use crate::{
     {debug, error, info, warn},
 };
 
-use core::convert::TryInto;
+use core::convert::{TryFrom, TryInto};
 use embedded_time::{
     duration::{Extensions, Milliseconds, Seconds},
     Instant,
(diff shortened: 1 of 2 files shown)


src/mqtt_client.rs
```diff
@@ -437,6 +450,12 @@ impl<'buf, TcpStack: TcpClientStack, Clock: embedded_time::Clock>
             return Ok(());
         }
 
+        if let Some(max) = &self.sm.context().max_qos {
+            if self.downgrade_qos && publish.qos > *max {
```
jordens, line 454: I think we'll also need to apply this to the `Will`.

jordens (commented):
Looks good.
Maybe this is a bit more than we should do.
e.g. mosquitto just errors early to the application:

https://github.com/eclipse/mosquitto/blob/f762a3fd1ced66b6cd32d8dc137f0523708dac1f/lib/actions.c#L54

src/mqtt_client.rs
ryan-summers: The will is set before the maximum QoS is known (it's sent as part of the CONNECT payload and we don't learn the max QoS until the CONNACK). I assume the broker will handle capping its QoS

jordens (approved):

PR #143, opened 2023-08-23
Refactoring configuration to builder semantics

ryan-summers:
This PR refactors the setup of the Minimq client into a single `Config` struct. This removes the need to do run-time checks on the client state and makes it simpler to see which buffers are being used for what. It also cuts down on the number of arguments needed in `Minimq::new()`


ryan-summers:
> Why not also move the broker to the config builder?

Was thinking about this as well. Seems reasonable to me :)

> And is it ok to keep the state_buffer empty, i.e. why is state_buffer treated differently from rx_buffer?

The `state_buffer` is only needed if the user is going to be publishing above `QoS::AtMostOnce`. There's no need for any state buffer if you're only sending simple AtMostOnce publications

jordens (approved):

PR #144, opened 2023-08-23
Refactoring publication to support zero-copy serialization

ryan-summers:
This PR is designed to allow an end user to provide a `Publication` message with a yet-to-be-serialized payload. This then allows them to serialize their payload _directly_ into the MQTT TX buffer, which eliminates the need for them serializing and then copying it over into the MQTT payload.

Example:
```rs
client.publish(Publication::new_deferred(|buf| settings.get(key, buf)));
```

### DIFF
--- a/src/publication.rs
+++ b/src/publication.rs
@@ -5,6 +5,47 @@ use crate::{
     ProtocolError, QoS, Retain,
 };
 
+pub trait ToPayload {
+    fn serialize(&self, buffer: &mut [u8]) -> Result<usize, ProtocolError>;
+}
+
+impl<'a> ToPayload for &'a [u8] {
+    fn serialize(&self, buffer: &mut [u8]) -> Result<usize, ProtocolError> {
+        if buffer.len() < self.len() {
+            return Err(ProtocolError::BufferSize);
+        }
+        buffer[..self.len()].copy_from_slice(self);
+        Ok(self.len())
+    }
+}
+
+impl<const N: usize> ToPayload for [u8; N] {
+    fn serialize(&self, buffer: &mut [u8]) -> Result<usize, ProtocolError> {
+        (&self[..]).serialize(buffer)
+    }
+}
+impl<const N: usize> ToPayload for &[u8; N] {
+    fn serialize(&self, buffer: &mut [u8]) -> Result<usize, ProtocolError> {
+        (&self[..]).serialize(buffer)
+    }
+}
+
+pub struct DeferedPayload<F: Fn(&mut [u8]) -> Result<usize, ProtocolError>> {
+    func: F,
+}
+
+impl<F: Fn(&mut [u8]) -> Result<usize, ProtocolError>> DeferedPayload<F> {
+    pub fn new(func: F) -> Self {
+        Self { func }
+    }
+}
+
+impl<F: Fn(&mut [u8]) -> Result<usize, ProtocolError>> ToPayload for DeferedPayload<F> {
+    fn serialize(&self, buffer: &mut [u8]) -> Result<usize, ProtocolError> {
+        (self.func)(buffer)
+    }
+}
+
 /// Builder pattern for generating MQTT publications.
 ///
 /// # Note
@@ -17,17 +58,17 @@ use crate::{
 /// It is expected that the user provide a topic either by directly specifying a publication topic
 /// in [Publication::topic], or by parsing a topic from the [Property::ResponseTopic] property
 /// contained within received properties by using the [Publication::reply] API.
-pub struct Publication<'a> {
+pub struct Publication<'a, P: ToPayload> {
     topic: Option<&'a str>,
     properties: Properties<'a>,
     qos: QoS,
-    payload: &'a [u8],
+    payload: P,
     retain: Retain,
 }
 
-impl<'a> Publication<'a> {
+impl<'a, P: ToPayload> Publication<'a, P> {
     /// Construct a new publication with a payload.
-    pub fn new(payload: &'a [u8]) -> Self {
+    pub fn new(payload: P) -> Self {
         Self {
             payload,
             qos: QoS::AtMostOnce,
@@ -121,7 +162,7 @@ impl<'a> Publication<'a> {
     /// # Returns
     /// The message to be published if a publication topic was specified. If no publication topic
     /// was identified, an error is returned.
-    pub fn finish(self) -> Result<Pub<'a>, ProtocolError> {
+    pub fn finish(self) -> Result<Pub<'a, P>, ProtocolError> {
         Ok(Pub {
             topic: Utf8String(self.topic.ok_or(ProtocolError::NoTopic)?),
             properties: self.properties,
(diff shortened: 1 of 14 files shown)


src/publication.rs
```diff
@@ -5,6 +5,47 @@ use crate::{
     ProtocolError, QoS, Retain,
 };
 
+pub trait ToPayload {
```
jordens, line 8: I don't get why it needs to be `Fn` (or `FnMut`) and not `FnOnce`.

And what about this?

```rust
enum Payload<'a, F: FnOnce(&mut [u8]) -> Result<usize, ()>> {
    Borrowed(&'a [u8]),
    Callback(F),
}
```

jordens: You can also just have the callback and then convert `Borrowed(src) => Callback(|buf| buf[..src.len()].copy_from_slice(src))`

ryan-summers: The problem with the enum is that it doesn't have a concrete type in the case where we're only using a `&[u8]` slice. We need the contrete type for `ReceivedPacket`, so there's a catch-22 there. I originally started iwth an enum and then ran into this.

jordens: Maybe split `Pub` packet into `IncomingPub` and `OutgoingPub`?

The idea of a callback for payload is pretty ambitious anyway. It requires fixing many additional variable length things (like the topic to respond to or properties) before knowing whether the payload generation/serialization even succeeds. We'd want to invert the serialization. We already do this for the fixed header with the remaining length varint. Three years ago we had the full reverse serialization (building the packet from the payload backwards).

ryan-summers: For example:
```
error[E0283]: type annotations needed
   --> src\mqtt_client.rs:480:40
    |
480 |                 let Ok(response_pub) = minimq::Publication::new(response.msg.as_bytes())
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type of the type parameter `F` declared on the struct `Publication`
...
490 |                 if client.publish(response_pub).is_err() {
    |                           ------- type must be known at this point
    |
    = note: multiple `impl`s satisfying `for<'a> _: FnOnce<(&'a mut [u8],)>` found in the following crates: `alloc`, `core`:
            - impl<A, F> FnOnce<A> for &F
              where A: Tuple, F: Fn<A>, F: ?Sized;
            - impl<A, F> FnOnce<A> for &mut F
              where A: Tuple, F: FnMut<A>, F: ?Sized;
            - impl<Args, F, A> FnOnce<Args> for std::boxed::Box<F, A>
              where Args: Tuple, F: FnOnce<Args>, A: Allocator, F: ?Sized;
note: required by a bound in `minimq::packets::OutgoingPub`
   --> C:\Users\rsummers\Documents\repositories\quartiq\minimq\src\packets.rs:77:34
    |
77  | pub struct OutgoingPub<'a, E, F: FnOnce(&mut [u8]) -> Result<usize, E>> {
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `OutgoingPub`
help: consider specifying the generic arguments
    |
480 |                 let Ok(response_pub) = minimq::Publication::<E, F>(response.msg.as_bytes())

```

tests/deferred_payload.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    assert!(matches!(
+        mqtt.client().publish(
+            DeferredPublication::new(|_buf| { Err("Oops!") })
+                .topic("data")
+                .qos(QoS::ExactlyOnce)
+                .finish()
+                .unwrap(),
+        ),
```
jordens, line 35: After this, when `publish()` errors, the app could instead send another Publication as the error message, right?

jordens (changes requested):
Looks very neat.

tests/deferred_payload.rs
ryan-summers: Yup! That's the intent. You can catch the error and use it to generate a response, this is the intended use case for miniconf

src/publication.rs
```diff
@@ -5,6 +5,55 @@ use crate::{
...
+    fn serialize(&self, buffer: &mut [u8]) -> Result<usize, Self::Error> {
+        if buffer.len() < self.len() {
+            return Err(());
+        }
+        buffer[..self.len()].copy_from_slice(self);
+        Ok(self.len())
+    }
+}
+
+impl<const N: usize> ToPayload for [u8; N] {
```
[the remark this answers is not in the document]
ryan-summers: Coersion doesn't appear to be working for the trait object because these are required for some of the unit tests:

Upon removal, various errors such as these are encountered:
```
note: required by a bound in `Pub`
   --> src\packets.rs:77:23
    |
77  | pub struct Pub<'a, P: crate::publication::ToPayload> {
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Pub`

error[E0277]: the trait bound `&[{integer}; 2]: ToPayload` is not satisfied
   --> src\packets.rs:384:66
    |
384 |         let message = MqttSerializer::pub_to_buffer(&mut buffer, &publish).unwrap();
    |                       -----------------------------              ^^^^^^^^ the trait `ToPayload` is not implemented for `&[{integer}; 2]`
    |                       |
    |                       required by a bound introduced by this call
    |
    = help: the trait `ToPayload` is implemented for `&'a [u8]`
note: required by a bound in `MqttSerializer::<'a>::pub_to_buffer`
   --> src\ser\mod.rs:144:29
    |
144 |     pub fn pub_to_buffer<P: crate::publication::ToPayload>(
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `MqttSerializer::<'a>::pub_to_buffer`

error[E0277]: the trait bound `&[{integer}; 2]: ToPayload` is not satisfied
   --> src\packets.rs:384:23
    |
384 |         let message = MqttSerializer::pub_to_buffer(&mut buffer, &publish).unwrap();
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `ToPayload` is not implemented for `&[{integer}; 2]`
    |
    = help: the trait `ToPayload` is implemented for `&'a [u8]`

error[E0277]: the trait bound `&[{integer}; 2]: ToPayload` is not satisfied
   --> src\packets.rs:384:23
    |
384 |         let message = MqttSerializer::pub_to_buffer(&mut buffer, &publish).unwrap();
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `ToPayload` is not implemented for `&[{integer}; 2]`
    |
    = help: the trait `ToPayload` is implemented for `&'a [u8]`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `minimq` (lib test) due to 16 previous errors
warning: build failed, waiting for other jobs to finish...
```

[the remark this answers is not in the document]
jordens: Ack!

PR #145, opened 2023-08-25
Updating will management

ryan-summers:
This PR refactors the `will` temporary storage such that the `will` is serialized at configuration time into a user-provided buffer. This means that the `Will` message itself does not need to live very long.

Pros and Cons:
* User has to provide yet another buffer at config time
* Libraries can leverage will memory provided by the user to construct customized wills.
* Will serialization only ever happens once


ryan-summers:
I spawned #146 to refactor, seems a bit of a bigger change than this PR was intending.

jordens (approved):

PR #146, opened 2023-08-28
Refactoring builder to accept a single buffer

ryan-summers:
This PR simplifies usability of the `Config` structure builder by only requiring a single user buffer.

This is a follow-on to #145 to simplify builder arguments.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,37 +3,57 @@ use core::convert::TryFrom;
 use embedded_time::duration::{Extensions, Milliseconds};
 use heapless::String;
 
-pub(crate) enum WillState<'a> {
-    BufferAvailable(&'a mut [u8]),
-    Serialized(SerializedWill<'a>),
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum BufferConfig {
+    /// Specify a buffer as having a minimum size in bytes.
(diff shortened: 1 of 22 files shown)


src/config.rs
```diff
@@ -81,27 +120,168 @@ impl<'a, Broker: crate::Broker> Config<'a, Broker> {
...
+                    Some((size, buf))
+                } else {
+                    None
+                }
+            })
+        {
+            let (head, tail) = data.split_at_mut(*size);
+            data = tail;
+            buffer.replace(head);
+        }
```
jordens, line 158: This is becoming more like an allocator.

jordens (approved):

src/config.rs
ryan-summers: It felt 100% like an allocator as I was writing it. I don't know if this is a good or bad thing.

In our case, we have an upper limit on the number of buffers we're splitting, and they never deallocate, so it's more like half an allocator.

ryan-summers: Perhaps we could steal memory from the `session_state` buffer for the will message instead of this heavy-handed approach of implementing a full allocator?

jordens: Let's keep it like this and see how it goes.

PR #147, opened 2023-08-31
Returning a public error type

ryan-summers:
This PR updates the `will()` API of the `ConfigBuilder` to return a public error type that is easily coerced. The `crate::ser::Error` is not publicly exported.

I also made it so these types are now publicly exported so users can easily match on them.


jordens (changes requested):
Let's Make these enums non_exhaustive for Denver proofing

jordens (approved):

PR #148, opened 2023-08-31
Updating will to use builder semantics

ryan-summers:
Updating the will to be a bit more usable with builder semantics. Ref https://github.com/quartiq/miniconf/pull/179#discussion_r1311439931

Also flagging an attempt to assign multiple wills as an error. This would potentially just lose allocated memory.

### DIFF
--- a/src/will.rs
+++ b/src/will.rs
@@ -90,7 +90,7 @@ impl<'a> Will<'a> {
     }
 
     /// Specify the will as a retained message.
-    pub fn set_retained(mut self) -> Self {
+    pub fn retained(mut self) -> Self {
         self.retained = Retain::Retained;
         self
     }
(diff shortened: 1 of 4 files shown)


src/will.rs
```diff
@@ -89,20 +89,19 @@ impl<'a> Will<'a> {
...
     }
 
-    /// Set the retained status of the will.
-    ///
-    /// # Args
-    /// * `retained` - Specifies the retained state of the will.
-    pub fn retained(&mut self, retained: Retain) {
-        self.retained = retained;
+    /// Specify the will as a retained message.
+    pub fn set_retained(mut self) -> Self {
```
jordens, line 93:
```suggestion
    pub fn retained(mut self) -> Self {
```
for builder pattern.

jordens (changes requested):

PR #149, opened 2023-09-05
Fixing pending write cache management

ryan-summers:
This PR fixes an issue I encountered with pending writes when integrating miniconf/minimq onto Stabilizer.

The cached pending write needs to use the `start` and `len` parameters when refreshing the cache indices, otherwise we accidentally cause malformed packets due to incorrect indices.


jordens (approved):
I think there were also spurious ci failures. Is that related?

ryan-summers:
Very potentially. I had noticed that as well.

PR #150, opened 2023-09-05
Refactoring auth data to not use heapless Strings, fixing build

ryan-summers:
This PR updates auth to:
1. Use the `ConfigBuilder` struct for specification to align with other start-up behaviors
2. Uses the `buffer` from the config to back the memory needed to cache the auth credentials, eliminating the need for heapless::String types (which impose run-time cost on users that don't need auth)
3. Excludes auth serialization without the expected feature flag to prevent accidentally providing insecure credentials.

CC @cnmozzie

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,4 +1,4 @@
-use crate::{will::SerializedWill, ProtocolError, Will};
+use crate::{types::Auth, will::SerializedWill, ProtocolError, Will};
 use core::convert::TryFrom;
 use embedded_time::duration::{Extensions, Milliseconds};
 use heapless::String;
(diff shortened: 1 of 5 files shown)


jordens (approved):
LGTM

In practice, is it easy to keep the username and password alive long enough to meet the borrowing lifetime requirement?

cnmozzie:
When I run `cargo build --features "unsecure"` it shows the error message:
```
Compiling minimq v0.7.0 (D:\code\rust\minimq)
error[E0308]: mismatched types
  --> src\config.rs:93:13
   |
93 |             user_name,
   |             ^^^^^^^^^ expected `&str`, found `&mut [u8]`
   |
   = note:      expected reference `&str`
           found mutable reference `&mut [u8]`

error[E0308]: mismatched types
  --> src\config.rs:94:13
   |
94 |             password,
   |             ^^^^^^^^ expected `&str`, found `&mut [u8]`
   |
   = note:      expected reference `&str`
           found mutable reference `&mut [u8]`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `minimq` (lib) due to 2 previous errors
```

cnmozzie:
> @cnmozzie Thanks for the heads up! I fixed CI so that it builds with --all-features now to prevent this in the future and cleaned up the code :)

Nice, can't wait to see the new release  :)

PR #152, opened 2023-09-08
Elevating ToPayload to FnOnce, adding ToPayload for &str

ryan-summers:
The `DeferredPayload` needs to be `FnOnce` (or `FnMut`) if we want the user to be able to mutate data within a deferred payload calculation.

This is necessary for things like i.e. `minireq`, where the user is going to be doing application-level logic within a `DeferredPayload` calculation, which may necessitate mutation of data and context.

I also implemented `ToPayload` for `&str` so you can now call `Publication::new("Hello")` without having to call `as_bytes()` on the str.

### DIFF
--- a/src/network_manager.rs
+++ b/src/network_manager.rs
@@ -148,7 +148,7 @@ where
 
     pub fn send_pub<P: crate::publication::ToPayload>(
         &mut self,
-        pub_packet: &Pub<'_, P>,
+        pub_packet: Pub<'_, P>,
     ) -> Result<&[u8], crate::PubError<TcpStack::Error, P::Error>> {
         // If there's an unfinished write pending, it's invalid to try to write new data. The
         // previous write must first be completed.
--- a/src/publication.rs
+++ b/src/publication.rs
@@ -7,13 +7,13 @@ use crate::{
 
 pub trait ToPayload {
     type Error;
-    fn serialize(&mut self, buffer: &mut [u8]) -> Result<usize, Self::Error>;
+    fn serialize(self, buffer: &mut [u8]) -> Result<usize, Self::Error>;
 }
 
 impl<'a> ToPayload for &'a [u8] {
     type Error = ();
 
-    fn serialize(&mut self, buffer: &mut [u8]) -> Result<usize, Self::Error> {
+    fn serialize(self, buffer: &mut [u8]) -> Result<usize, Self::Error> {
         if buffer.len() < self.len() {
             return Err(());
         }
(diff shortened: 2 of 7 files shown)


src/network_manager.rs
```diff
@@ -148,7 +148,7 @@ where
 
     pub fn send_pub<P: crate::publication::ToPayload>(
         &mut self,
-        pub_packet: &Pub<'_, P>,
+        pub_packet: Pub<'_, P>,
```
jordens, line 151: Why this?

src/publication.rs
```diff
@@ -22,34 +22,42 @@ impl<'a> ToPayload for &'a [u8] {
...
     type Error = ();
 
-    fn serialize(&self, buffer: &mut [u8]) -> Result<usize, ()> {
+    fn serialize(&mut self, buffer: &mut [u8]) -> Result<usize, ()> {
         (&self[..]).serialize(buffer)
     }
 }
 
-pub struct DeferredPublication<E, F: Fn(&mut [u8]) -> Result<usize, E>> {
+pub struct DeferredPublication<E, F: FnMut(&mut [u8]) -> Result<usize, E>> {
```
jordens, line 48: Why not fnonce? We only ever call it once and we can consume the closure.

ryan-summers: This is needed by users like `minireq` (ref https://github.com/quartiq/minireq/pull/16)) -  The `FnMut` states that the _user_ can mutate objects within the closure. We have no knowledge of what the closure is, so the most relaxed trait bound would be `FnMut`

src/network_manager.rs
ryan-summers: This eases ownership semantics because of the new FnMut requiring mut access to the `pub_packet`

ryan-summers: If we didn't take ownership, we'd have to use `&mut`, which causes all kinds of angry compiler declarations about `&mut self` and `&mut pub_packet` ownership lifetimes.

src/publication.rs
jordens: Taking a `FnOnce` argument does not preclude it from mutate objects within its capture! That's not how these work. An `FnOnce` can mutate things just fine. Yes, `FnMut` is more relaced than `Fn`. But from your description and from how we use it, an `FnOnce` will do. Every `FnMut` is also `FnOnce`.

ryan-summers: Indeed, `FnOnce` does indeed look much nicer. I didn't realize the correlation between `Fn`, `FnMut`, `FnOnce` and `&self`, `&mut self`, and `self` until now, thanks for pointers in the right direction.

src/network_manager.rs
ryan-summers: With FnOnce, we now require ownership anyways

jordens: Correct.
I think you could in principle still do fine with a &Mut and mem::swap before calling func but it doesn't seem sensible.

But then (if we consume self) I'd try avoiding `impl topayload for [u8;N]` and instead only borrow. Otherwise I'd worry about copies (small arrays are copy).

jordens (approved):

PR #155, opened 2023-11-23
Handle bad length in header

madmo:
Prevents a panic in `fn receive_buffer` if a packet with a too big length is received.

### DIFF
--- a/src/de/packet_reader.rs
+++ b/src/de/packet_reader.rs
@@ -27,7 +27,11 @@ impl<'a> PacketReader<'a> {
             self.read_bytes + 1
         };
 
-        Ok(&mut self.buffer[self.read_bytes..end])
+        if end <= self.buffer.len() {
+            Ok(&mut self.buffer[self.read_bytes..end])
+        } else {
+            Err(Error::MalformedPacket)
+        }
     }
 
     pub fn commit(&mut self, count: usize) {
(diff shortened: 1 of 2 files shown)


src/de/packet_reader.rs
```diff
@@ -55,6 +55,12 @@ impl<'a> PacketReader<'a> {
             }
         }
 
+        if let Some(packet_length) = self.packet_length {
```
ryan-summers, line 58: I'm not following what went wrong and why this fixes it. Can you add a unit test case that shows what you're observing and also demonstrates that this is now fixed?

ryan-summers (changes requested):
Hey there! Thanks for the PR :)

I have some minor questions since I don't fully understand the issue.

src/de/packet_reader.rs
madmo: Thanks for getting back to me. Pushed this out at the end of day, which was not a good idea.
Updated the pull request to correctly fix the issue, and also with a small testcase.

ryan-summers (changes requested):
It looks like this is protecting against an issue where the packet-header of the MQTT datagram may be corrupt and indicate a much larger size than we support, which we obviously should not use to index into `self.buffer`, so this is definitely a bug. Thanks for the fix here!

Can you please:
- [ ] Add an entry to CHANGELOG.md indicating this defect
- [ ] Fix the style formatting?

ryan-summers (approved):
Thanks! This looks like a nice fix. I can cut a patch release out for this as well.

PR #157, opened 2024-03-19
bump embedded-nal

jordens:


jordens:
more work to be done around stack mut/refs.

ryan-summers (approved):

PR #184, opened 2026-04-27
fix ping strategy

kkocdko:
Work in progress.

Currently the ping strategy seems not follow spec.

https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#:~:text=maximum%20time%20interval

> The Keep Alive is a Two Byte Integer which is a time interval measured in seconds. It is the maximum time interval that is permitted to elapse between the point at which the Client finishes transmitting one MQTT Control Packet and the point it starts sending the next. It is the responsibility of the Client to ensure that the interval between MQTT Control Packets being sent does not exceed the Keep Alive value. If Keep Alive is non-zero **and in the absence of sending any other MQTT Control Packets**, the Client MUST send a PINGREQ packet.

TODO: what does other implement do?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ heapless = { version = "0.9.1" }
 log = { version = "0.4", default-features = false }
 embedded-io-async = "0.7"
 embedded-nal-async = "0.9"
-embassy-futures = "0.1.2"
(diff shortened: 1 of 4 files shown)


kkocdko:
To issue #186

PR #185, opened 2026-04-27
no connector

jordens:
- **is_publish_quiescent() and ToPayload trait cleanup**
- **mqtt: split explicit connect from bounded poll**
- **raise retained write tracing to debug**
- **remove Connector, Broker, let user resolve and connect TCP**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Edition 2024, MSRV 1.85
 * Redesign and rebase onto async (`embedded-io-async`). This breaks most of the API surface.
   There is no sync compat layer.
-* Broker session handling is now explicit: `Event::Connected` means a fresh broker session,
-  `Event::Reconnected` means the broker resumed the existing session.
+* Broker session handling is now explicit: `ConnectEvent::Connected` means a fresh broker session,
+  `ConnectEvent::Reconnected` means the broker resumed the existing session.
 * Caller-owned packet buffers are now documented as `rx` and `tx`. `rx` holds one inbound packet;
   `tx` is the shared outbound encode/replay arena.
 * Plain MQTT username/password auth is always available. The crate-local `unsecure` feature was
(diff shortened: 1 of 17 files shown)


kkocdko:
Humm, curious how to make keep_alive work? starts from this no-connector refactor, even the basic embassy TcpSocket (has read_ready and write_ready) will cause idle disconnect. Did I do something wrong?


```rust

let mut socket = embassy_net::tcp::TcpSocket::new(stack, tcp_rx, tcp_tx);
let host = MQTT_HOST;
let port = MQTT_PORT;
info!("wanna to query domain: {}", host);
socket.connect(IpEndpoint { addr: "35.172.255.228".parse().unwrap(), port }).await.unwrap(); // emqx
info!("connected with ip without dns query");

let config = ConfigBuilder::new(Buffers::new(rx, tx))
    .keepalive_interval(5)
    .session_expiry_interval(3600)
    .auth(MQTT_AUTH_USER, MQTT_AUTH_PASSWORD.as_bytes())
    .unwrap()
    .client_id("the-client-id-from-minimq")
    .unwrap();
let mut session = Session::new(config);
// session.publish(reply.qos(QoS::AtLeastOnce)).await.unwrap();
let topic_command = "device/the-id/command";
let topic_command = TopicFilter::new(topic_command)
    .options(SubscriptionOptions::default().maximum_qos(QoS::AtLeastOnce)); // no qos 2
let topic_event = "device/the-id/event";
if let Err(_) = session.connect(socket).await {
    return Err(MqttUpError::Disconnected);
}
session
    .subscribe(&[topic_command], &[])
    .await
    .or(Err(MqttUpError::Mqtt))?;
info!("mqtt subscribed");
loop {
    // session
    match session.poll().await {
        Ok(Event::Inbound(message)) => {
            info!(
                "mqtt received topic: {}, payload: {}, qos: {}",
                message.topic(),
                message.payload(),
                message.qos() as u8
            );
            session
                .publish(
                    Publication::new(topic_event, "the-publish-payload").qos(QoS::AtMostOnce), // qos 0
                )
                .await
                .or(Err(MqttUpError::Mqtt))?;
            // if let Some(_reply) = message.reply("ack") {}
        }
        Ok(Event::Idle) => {}
        Err(e) => {
            let s = alloc::format!("{:?}", e);
            warn!("mqtt error: {}", s);
            return Err(MqttUpError::Mqtt);
        }
    }
}
```


IMO, depends on embassy-time with_deadline is acceptable.

PR #187, opened 2026-04-27
fix ping

### LINKED ISSUE #186: Omit unnecessary keep alive ping
https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#:~:text=maximum%20time%20interval

> The Keep Alive is a Two Byte Integer which is a time interval measured in seconds. It is the maximum time interval that is permitted to elapse between the point at which the Client finishes transmitting one MQTT Control Packet and the point it starts sending the next. It is the responsibility of the Client to ensure that the interval between MQTT Control Packets being sent does not exceed the Keep Alive value. If Keep Alive is non-zero and in the absence of sending any other MQTT Control Packets, the Client MUST send a PINGREQ packet.

Other implements:

[eclipse/paho](github.com/eclipse-paho/paho.mqtt.c) do these well, in it's MQTTProtocolClient.c:

https://github.com/eclipse-paho/paho.mqtt.c/blob/v1.3.16/src/MQTTProtocolClient.c#L748-L752

Only send PINGREQ when `difftime(now, client->net.lastSent) >= client->keepAliveInterval`, and, send any packet successfully will update `lastSent`.

@jordens  Seems that you're doing a big refactor, my PR got outdated, hope this feature to be considered.

jordens:
- **fix ping timeout updates**
- **outbound: extract common path**

### DIFF
--- a/src/mqtt_client/core.rs
+++ b/src/mqtt_client/core.rs
@@ -27,6 +27,13 @@ enum ReadMode {
     Blocking,
 }
 
+#[derive(Copy, Clone)]
+enum FlushedPacket {
+    Control(ControlAction),
+    Release(u16),
+    Retained(u16),
+}
+
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub(super) enum ConnectionState {
     Disconnected,
@@ -70,8 +77,8 @@ impl RuntimeState {
     }
 
     pub(super) fn note_outbound_activity(&mut self, now: Instant) {
-        self.next_ping = (self.keepalive_interval.as_secs() != 0)
-            .then_some(now + self.keepalive_interval)
+        self.next_ping =
+            (self.keepalive_interval.as_secs() != 0).then_some(now + self.keepalive_interval)
     }
 }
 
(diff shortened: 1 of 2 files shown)


kkocdko (approved):
I took a careful look, and it looks wonderful, thanks!

src/mqtt_client/core.rs
```diff
@@ -815,20 +817,11 @@ impl<'buf> Core<'buf> {
...
-                    {
-                        Ok(count) => count,
-                        Err(err) => {
-                            self.handle_disconnect();
-                            return Err(err);
-                        }
-                    };
+                    let count = self.disconnect_on_error({
+                        let packet = self.session.outbound.retained_packet(step.offset, step.len);
+                        Self::write_step_bytes(connection, "Retained packet", &packet[written..])
```
kkocdko: I found that the `log::log!("content")` will be optimized in release build, but function call with static string will not, like `Self::write_step_bytes(connection, "Retained packet", xxx)`.

```
kkocdko@klp1:~/misc/code/kmbed/aio-esp32c3$ cat target/riscv32imc-unknown-none-elf/release/aio-esp32c3 | grep --binary "Flushing PUBREL packet packet_id"
kkocdko@klp1:~/misc/code/kmbed/aio-esp32c3$ 
kkocdko@klp1:~/misc/code/kmbed/aio-esp32c3$ cat target/riscv32imc-unknown-none-elf/release/aio-esp32c3 | grep --binary "Retained packet"
grep: (standard input): binary file matches
kkocdko@klp1:~/misc/code/kmbed/aio-esp32c3$ 
```

However, may not a big problem.

jordens: Oh yeah. That's bad.

kkocdko (approved):

PR #202, opened 2026-06-18
Split transport IO out of `Session` into a `Connection` handle

Ddystopia:
## Motivation

`Session<'buf, IO>` owns the transport for the whole session lifetime, which bakes a single `IO`
type **and lifetime** into the session type. That makes it impossible to reuse one `Session` across
reconnects when the transport's lifetime differs between connections - most painfully over TLS,
where the record buffers are borrowed for the connection's lifetime and so have to outlive the
session. In practice that forces a fresh allocation (a leak) per connection: the old
`tls_public_broker` example `Box::leak`s a new record-buffer pair every time it connects.

## What changed

`Session` loses its `IO` generic. `Session::connect::<IO>(io)` now returns a
**`Connection<'_, 'buf, IO>`** handle that owns the transport and borrows the session:

```rust
let mut session = Session::new(config);          // no IO in the type
loop {
    let io = open_transport().await?;            // a fresh IO, any lifetime
    let mut conn = session.connect(io).await?;   // handle owns io, borrows session
    let op = conn.publish(/* … */).await?;
    while conn.is_pending(&op) { conn.poll().await?; }
    // conn drops here -> transport (and its buffer borrows) released -> reuse next loop
}
```

- All network operations (`drive`/`poll`/`recv`/`publish`/`subscribe`/`unsubscribe`/`disconnect`)
  live on the handle; session-state queries (`is_pending`, `can_publish`, …) are reachable through
  `Deref` to `Session`. `connect_event()` exposes the `Connected`/`Reconnected` result.
- **`mem::forget`-safe by construction.** The handle may be dropped *or* `mem::forget`-ed at any
  time (e.g. to release a TLS record-buffer borrow without running `Drop`), so correctness can't
  depend on `Drop`. All state carried across reconnects - in-flight QoS replay, keepalive timers,
  the inbound packet reader - is reset at the **start of every `connect()`** instead of in a
  disconnect/`Drop` path. The old `handle_disconnect()` is split accordingly: transport-drop -> the
  handle's ownership; replay/timer/reader resets -> `connect()`; and its liveness latch -> a
  `connected` flag set at the sites that detect a fatal disconnect.
- **`is_connected()` moves to the `Connection` handle** and reports *per-connection* liveness: it
  latches `false` the moment any operation observes a transport- or protocol-level disconnect
  (broker `DISCONNECT`, transport error, keepalive timeout) or a graceful `disconnect`, after which
  further operations fail fast with `Error::Disconnected` instead of driving a dead transport.
- `Connection::disconnect`/`disconnect_with` take `&mut self` so a cancelled disconnect can be
  retried.

## Drop vs. graceful disconnect

Dropping (or `mem::forget`-ing) the handle is an **ungraceful** MQTT close: no `DISCONNECT` packet
is sent (a sync `Drop` cannot perform the async write), so the broker sees an abnormal disconnect
and will publish the configured Will, if any. Reconnecting afterwards resumes the broker session
(`CONNECT` `clean_start=false` -> `Reconnected`). For a clean shutdown - `DISCONNECT` sent, Will
suppressed - call `conn.disconnect().await` before dropping. This is documented on `Connection`,
`connect`, and `disconnect`, and covered by a test asserting a bare drop emits no `DISCONNECT` and
that the next `connect` resumes.

## Showcase

The `tls_public_broker` example is reworked to **own one pair of TLS record buffers and reuse them
across a reconnect** - connect, round-trip, drop the handle, reconnect reusing the same buffers -
with no `Box::leak` and no per-connection allocation. The previous API could not express this.

## Testing

Whole suite green: **68** unit tests, **59** `async_client` integration tests, **2** `real_broker`
tests, **3** doctests - all ported to the handle API, plus new regression tests for the disconnect
latch and the ungraceful-drop/resume behavior.

## Breaking changes

- `Session` loses its `IO` type parameter; `Session::connect()` returns a `Connection` handle.
- Network operations move from `Session` to the handle (queries remain via the handle's `Deref`).
- `is_connected()` moves from `Session` to `Connection`.
- `Connection::disconnect`/`disconnect_with` take `&mut self`.

The README, all examples, and the test suite are updated to match. See the `CHANGELOG.md`
`[Unreleased]` entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased](https://github.com/quartiq/minimq/compare/v0.12.0...HEAD)
+
+## Changed
+
+* **Breaking:** `Session` no longer carries the transport `IO` type parameter. `Session::connect()`
+  now returns a `Connection` handle that owns the transport and borrows the session; the network
+  operations (`drive`/`poll`/`recv`/`publish`/`subscribe`/`unsubscribe`/`disconnect`) move onto the
+  handle, while session-state queries remain reachable through its `Deref` to `Session`. This lets a
+  single `Session` be reused across reconnects with transports of different types or buffer
+  lifetimes — for example reusing TLS record buffers instead of leaking a fresh pair per connection.
+* **Breaking:** `is_connected()` moves from `Session` to the `Connection` handle and now reports
+  *per-connection* liveness: it latches `false` as soon as any operation observes a transport- or
+  protocol-level disconnect (broker `DISCONNECT`, transport error, keepalive timeout) or a graceful
+  `disconnect`, after which further operations on the handle fail fast with `Error::Disconnected`
+  instead of driving the dead transport.
+
 ## [v0.12.0](https://github.com/quartiq/minimq/compare/v0.11.2..v0.12.0) - 2026-05-24
 
 ## Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * **Breaking:** `Session` no longer carries the transport `IO` type parameter. `Session::connect()`
   now returns a `Connection` handle that owns the transport and borrows the session; the network
   operations (`drive`/`poll`/`recv`/`publish`/`subscribe`/`unsubscribe`/`disconnect`) move onto the
-  handle, while session-state queries remain reachable through its `Deref` to `Session`. This lets a
-  single `Session` be reused across reconnects with transports of different types or buffer
-  lifetimes — for example reusing TLS record buffers instead of leaking a fresh pair per connection.
+  handle.
 * **Breaking:** `is_connected()` moves from `Session` to the `Connection` handle and now reports
   *per-connection* liveness: it latches `false` as soon as any operation observes a transport- or
   protocol-level disconnect (broker `DISCONNECT`, transport error, keepalive timeout) or a graceful
--- a/examples/tls_public_broker.rs
+++ b/examples/tls_public_broker.rs
@@ -51,7 +51,7 @@ async fn run() -> Result<(), Box<dyn StdError>> {
     let mut subscriber = Session::new(
         ConfigBuilder::from_buffer(&mut sub_storage, 1024)?.auth(USERNAME, PASSWORD.as_bytes())?,
     );
-    subscriber
+    let mut subscriber = subscriber
         .connect(connect_tls(BROKER_HOST, BROKER_PORT).await?)
         .await?;
     let sub = subscriber
@@ -65,7 +65,7 @@ async fn run() -> Result<(), Box<dyn StdError>> {
     let mut publisher = Session::new(
         ConfigBuilder::from_buffer(&mut pub_storage, 1024)?.auth(USERNAME, PASSWORD.as_bytes())?,
     );
-    publisher
+    let mut publisher = publisher
         .connect(connect_tls(BROKER_HOST, BROKER_PORT).await?)
         .await?;
     let pub_ = publisher
--- a/examples/tls_public_broker.rs
+++ b/examples/tls_public_broker.rs
@@ -1,11 +1,17 @@
 //! TLS connectivity example for `embedded-tls`.
 //!
-//! Use this as a TLS transport example. Bounded/cooperative session driving is done by wrapping
-//! the cancel-safe blocking `Session::poll()` in an external timeout at the call site.
+//! It also showcases the connection-handle API: the transport lives in the
+//! [`Connection`](minimq::Connection) handle returned by [`Session::connect`], not in the
+//! `Session`, so the TLS record buffers only need to live as long as one connection. A single pair
+//! of record buffers is owned here and re-lent to every connection, so one allocation serves every
+//! reconnect.
 
 use embedded_io_adapters::tokio_1::FromTokio;
 use embedded_tls::{Aes128GcmSha256, TlsConfig, TlsConnection, TlsContext, UnsecureProvider};
-use minimq::{ConfigBuilder, Publication, QoS, Session, TopicFilter};
+use minimq::{
+    ConfigBuilder, ConnectEvent, Connection, Error, Io, PubError, Publication, QoS, Session,
+    TopicFilter,
+};
 use std::error::Error as StdError;
 use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::net::TcpStream;
@@ -17,17 +23,28 @@ const PASSWORD: &str = "public";
 const TLS_READ_RECORD_BUFFER_LEN: usize = 16_640;
 const TLS_WRITE_RECORD_BUFFER_LEN: usize = 4_096;
 
-async fn connect_tls(
+/// Number of successful round-trips to perform before exiting. A real long-running client would
+/// loop forever; we stop after a few so the example terminates while still reconnecting (and so
+/// reusing the record buffers) more than once.
+const ROUND_TRIPS: usize = 2;
+
+/// Open a TLS connection over the caller-owned record buffers.
+///
+/// The returned connection borrows `read_record`/`write_record` for its own lifetime `'a`, not
+/// `'static`, which is what lets the caller reuse the buffers once the connection is dropped.
+async fn connect_tls<'a>(
     host: &str,
     port: u16,
-) -> Result<TlsConnection<'static, FromTokio<TcpStream>, Aes128GcmSha256>, Box<dyn StdError>> {
+    read_record: &'a mut [u8],
+    write_record: &'a mut [u8],
+) -> Result<TlsConnection<'a, FromTokio<TcpStream>, Aes128GcmSha256>, Box<dyn StdError>> {
     let config = TlsConfig::new()
         .with_server_name(host)
         .enable_rsa_signatures();
     let mut tls = TlsConnection::new(
         FromTokio::new(TcpStream::connect((host, port)).await?),
-        Box::leak(Box::new([0u8; TLS_READ_RECORD_BUFFER_LEN])),
-        Box::leak(Box::new([0u8; TLS_WRITE_RECORD_BUFFER_LEN])),
+        read_record,
+        write_record,
     );
     let mut provider = UnsecureProvider::new::<Aes128GcmSha256>(rand::rngs::OsRng);
     tls.open(TlsContext::new(&config, &mut provider)).await?;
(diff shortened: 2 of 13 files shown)


jordens:
Thanks for the contribution!
This identifies a real issue but picks too large a solution.

The real issue is not “MQTT needs a connection handle API.” It is: `Session<'buf, IO>` owns `Option<IO>`, and some embedded TLS transports borrow connection-local buffers. If minimq drops IO internally on disconnect/error, the caller cannot explicitly recover or release/recycle the TLS object and its borrowed buffers at the right phase boundary.

The handle proposal fixes one class of lifetime pressure by moving IO out of `Session`, but it spreads the live transport lifetime through the public MQTT API. It also keeps monomorphization in practice because methods become generic over IO; and if users truly need dynamic transport switching, `dyn Io` is already a plausible escape hatch. Different concrete IO types across one durable MQTT session do not seem important enough to justify redesigning the common API.

The cleaner architecture is:

Session = durable MQTT state
IO = current live transport
TLS/socket buffers = transport-layer resources owned/lent by the app

Session can keep Option<IO>. The missing API is explicit detachment, not a new public connection model.

Proposal:

```rust
pub async fn disconnect(&mut self) -> Result<Option<IO>, Error<IO::Error>>;
pub async fn disconnect_with(
    &mut self,
    disconnect: Disconnect<'_>,
) -> Result<Option<IO>, Error<IO::Error>>;

pub fn take_connection(&mut self) -> Option<IO>;
```

Rationale:

- disconnect() sends MQTT DISCONNECT, flushes, then returns the detached IO.
- Ok(None) keeps current idempotent “already disconnected” behavior.
- take_connection() gives an explicit ungraceful detach path for dead/abandoned transports.
- connect(io) can remain Error::NotReady if an IO is still attached.
- The TLS-buffer lifecycle becomes simple: disconnect/take old IO, drop it, reuse buffers, build new TLS IO, reconnect.
- Cancellation safety is preserved if disconnect() writes through self.connection.as_mut() and only take()s after successful completion.

This is a smaller, more conservative change. It preserves minimq’s current model, avoids leaking TLS-specific lifetime mechanics into MQTT, and still solves the practical buffer-reuse problem.

Ddystopia:
@jordens Look, the issue isn't in drop of the IO, but in the borrow-checker. The lifetime of the buffer is a type too. The borrow-checker doesn't know that `pub fn take_connection(&mut self) -> Option<IO>;` actually moved out the `IO` from the inside. If you don't want a contribution then close a PR, I won't contribute anything ever if you'll be spitting AI nonsense without verifying it, I'm sorry for harsh words.

jordens:
Looky looky my friend!
The borrow checker absolutely tracks that moving the `IO` out releases the borrow! That is exactly why `Option::take()` is the relevant operation here.
In your use case and the (expanded) TLS example the `Session<'buf, IO>` stays compatible across reconnects.
What you need is something that returns the IO without dropping it so the borrow ends!
If you have a concrete reduced example where `take_connection(); drop(io); make_new_tls_from_same_buffers(); session.connect(new_io)` still fails, that would be valuable. Without that, your handle rewrite is unmotivated.
Let whoever/whatever wrote your commits read the proposal and give it a try. Assure me that you wrote that PR and commits manually and you'll get a human response.

Ddystopia:
So you want to push the verification of your (probably) ideas on me, fine. Is this skeleton ([playground]) enough for you, or you want me to push a fork of minimq with `take_connection`?

Further notes: you might next say that I can feed the `io1` into the next connect (different from your request) and it will compile, but won't work: new transport has to be created, closed transport can't be reused. 

> Assure me that you wrote that PR and commits manually and you'll get a human response.

The changes are pure boilerplate, just moving methods around, I didn't write it manually, why write boilerplate manually? Or do you suggest that responding without verifying anything is equivalent to the design process?

cc @jordens 

[playground]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&code=struct+Session%3CIO%3E+%7B%0A++++connection%3A+Option%3CIO%3E%2C%0A%7D%0A%0Aimpl%3CIO%3E+Session%3CIO%3E+%7B%0A++++fn+new%28%29+-%3E+Self+%7B%0A++++++++Self+%7B+connection%3A+None+%7D%0A++++%7D%0A%0A++++fn+connect%28%26mut+self%2C+io%3A+IO%29+%7B%0A++++++++self.connection+%3D+Some%28io%29%3B%0A++++%7D%0A%0A++++fn+take_connection%28%26mut+self%29+-%3E+Option%3CIO%3E+%7B%0A++++++++self.connection.take%28%29%0A++++%7D%0A%7D%0A%0Astruct+BorrowingIo%3C%27a%3E+%7B%0A++++_read_record%3A+%26%27a+mut+%5Bu8%5D%2C%0A++++_write_record%3A+%26%27a+mut+%5Bu8%5D%2C%0A%7D%0A%0Afn+make_new_tls_from_same_buffers%3C%27a%3E%28%0A++++read_record%3A+%26%27a+mut+%5Bu8%5D%2C%0A++++write_record%3A+%26%27a+mut+%5Bu8%5D%2C%0A%29+-%3E+BorrowingIo%3C%27a%3E+%7B%0A++++BorrowingIo+%7B%0A++++++++_read_record%3A+read_record%2C%0A++++++++_write_record%3A+write_record%2C%0A++++%7D%0A%7D%0A%0Afn+main%28%29+%7B%0A++++let+mut+read_record+%3D+%5B0u8%3B+64%5D%3B%0A++++let+mut+write_record+%3D+%5B0u8%3B+16%5D%3B%0A%0A++++let+mut+session%3A+Session%3CBorrowingIo%3C%27_%3E%3E+%3D+Session%3A%3Anew%28%29%3B%0A%0A++++let+io1+%3D+make_new_tls_from_same_buffers%28%26mut+read_record%2C+%26mut+write_record%29%3B%0A++++session.connect%28io1%29%3B%0A%0A++++let+io1+%3D+session.take_connection%28%29.unwrap%28%29%3B%0A++++drop%28io1%29%3B%0A%0A++++let+io2+%3D+make_new_tls_from_same_buffers%28%26mut+read_record%2C+%26mut+write_record%29%3B%0A++++session.connect%28io2%29%3B%0A%0A++++let+_+%3D+session.take_connection%28%29%3B%0A%7D%0A

Ddystopia:
Note: I don't need your human response. I need your human verification. Please keep the middle man between us, just run the code and fact check before saying.

src/mqtt_client/session/mod.rs
```diff
@@ -145,3 +151,130 @@ where
...
+    pub(super) event: ConnectEvent,
+}
+
+impl<'buf, IO> core::ops::Deref for Connection<'_, 'buf, IO> {
+    type Target = Session<'buf>;
+
+    fn deref(&self) -> &Self::Target {
+        self.session
+    }
+}
```
jordens: No Deref magic. Do `pub fn session(&self) -> &Session`.

src/mqtt_client/session/mod.rs
```diff
@@ -44,12 +45,15 @@ pub struct Session<'buf, IO> {
     auth: Option<Auth<'buf>>,
     session_expiry_interval: u32,
     downgrade_qos: bool,
+    /// Whether the transport currently owned by an outstanding [`Connection`] is still usable.
+    /// Set true by a successful [`connect`](Self::connect) and latched false the moment any
+    /// operation observes a fatal transport- or protocol-level disconnect. State carried to the
+    /// next connection is reset in `connect`, not here, so the latch does not depend on the
+    /// handle's `Drop` running.
+    connected: bool,
```
jordens: This needs to be `live: bool` on `Connection`. Otherwise dropping `Connection` leaves `Session` `connected`.

The reset of transport-local session state belongs in `Session`, but the live/dead latch belongs in Connection.

examples/tls_public_broker.rs
```diff
@@ -1,11 +1,17 @@
 //! TLS connectivity example for `embedded-tls`.
 //!
-//! Use this as a TLS transport example. Bounded/cooperative session driving is done by wrapping
-//! the cancel-safe blocking `Session::poll()` in an external timeout at the call site.
+//! It also showcases the connection-handle API: the transport lives in the
+//! [`Connection`](minimq::Connection) handle returned by [`Session::connect`], not in the
```
jordens: Do not explain what it doesn't do. That's AI stink.

examples/tls_public_broker.rs
```diff
@@ -17,17 +23,28 @@ const PASSWORD: &str = "public";
 const TLS_READ_RECORD_BUFFER_LEN: usize = 16_640;
 const TLS_WRITE_RECORD_BUFFER_LEN: usize = 4_096;
 
-async fn connect_tls(
+/// Number of successful round-trips to perform before exiting. A real long-running client would
+/// loop forever; we stop after a few so the example terminates while still reconnecting (and so
+/// reusing the record buffers) more than once.
+const ROUND_TRIPS: usize = 2;
```
jordens: That makes this example complicated and harder to read. Can we stick to the simple existing one? The AI comment lecturing is annoying! If all you need is removing the IO type from Session, we don't need an example to show that this PR does it.

examples/tls_public_broker.rs
```diff
@@ -17,17 +23,28 @@ const PASSWORD: &str = "public";
...
+
+/// Open a TLS connection over the caller-owned record buffers.
+///
+/// The returned connection borrows `read_record`/`write_record` for its own lifetime `'a`, not
+/// `'static`, which is what lets the caller reuse the buffers once the connection is dropped.
```
jordens: Don't bother explaining what it doesn't do.

src/mqtt_client/session/mod.rs
```diff
@@ -145,3 +151,130 @@ where
...
+/// suppressed — call [`disconnect`](Self::disconnect) (or [`disconnect_with`](Self::disconnect_with))
+/// before dropping. Reconnecting afterwards with [`Session::connect`](Session::connect) resumes the
+/// broker session via `CONNECT` `clean_start=false`; see [`ConnectEvent`](crate::ConnectEvent).
+pub struct Connection<'a, 'buf, IO> {
+    pub(super) session: &'a mut Session<'buf>,
+    pub(super) io: IO,
+    pub(super) event: ConnectEvent,
+}
+
+impl<'buf, IO> core::ops::Deref for Connection<'_, 'buf, IO> {
```
jordens: I think you'll want `into_inner() -> IO`. Are you OK with only supprting drop for it?

src/mqtt_client/session/mod.rs
```diff
@@ -145,3 +151,130 @@ where
...
+    /// Gracefully close the transport with `DISCONNECT`.
+    ///
+    /// This is the graceful counterpart to simply dropping the handle: it sends the MQTT
+    /// `DISCONNECT` so the broker closes cleanly and suppresses the Will. Just dropping (or
+    /// [`mem::forget`](core::mem::forget)-ing) the handle skips this and is treated by the broker
+    /// as an abnormal disconnect.
+    ///
+    /// Takes `&mut self` rather than consuming the handle, so a cancelled disconnect can be
+    /// retried. After a successful disconnect the handle's transport is dead; drop it (or
+    /// `mem::forget` it) to release the session for a fresh `connect`.
```
jordens: Repetitive rationalizing and unrelated information. It leaks thought development process thoughts into the docs. Please remove this.

src/mqtt_client/session/mod.rs
```diff
@@ -73,12 +76,15 @@ where
...
         }
     }
 
-    /// Return whether the MQTT session is currently established.
-    pub fn is_connected(&self) -> bool {
-        self.connection.is_some()
+    /// Latch the current transport as disconnected. Called at every site that observes a fatal
+    /// transport- or protocol-level failure. Only flips the liveness flag; state needed for the
+    /// next connection is reset in `connect`.
+    pub(super) fn mark_disconnected(&mut self) {
```
jordens: That's not sufficient.
We need to call something like `self.session.reset_transport_state_for_reconnect();` in Connection like the old `handle_disconnect()`. Otherwise e.g. `can_publish()` can be true while `Connection` has been dropped.

src/mqtt_client/session/mod.rs
```diff
@@ -145,3 +151,130 @@ where
...
+        P: ToPayload,
+    {
+        self.session.publish(&mut self.io, publication).await
+    }
+
+    /// Gracefully close the transport with `DISCONNECT`.
+    ///
+    /// This is the graceful counterpart to simply dropping the handle: it sends the MQTT
+    /// `DISCONNECT` so the broker closes cleanly and suppresses the Will. Just dropping (or
+    /// [`mem::forget`](core::mem::forget)-ing) the handle skips this and is treated by the broker
```
jordens: What's with this repeated riffing on about mem::forget()?

CHANGELOG.md
```diff
@@ -6,6 +6,22 @@ All notable changes to this project will be documented in this file.
...
+  now returns a `Connection` handle that owns the transport and borrows the session; the network
+  operations (`drive`/`poll`/`recv`/`publish`/`subscribe`/`unsubscribe`/`disconnect`) move onto the
+  handle, while session-state queries remain reachable through its `Deref` to `Session`. This lets a
+  single `Session` be reused across reconnects with transports of different types or buffer
+  lifetimes — for example reusing TLS record buffers instead of leaking a fresh pair per connection.
+* **Breaking:** `is_connected()` moves from `Session` to the `Connection` handle and now reports
+  *per-connection* liveness: it latches `false` as soon as any operation observes a transport- or
+  protocol-level disconnect (broker `DISCONNECT`, transport error, keepalive timeout) or a graceful
+  `disconnect`, after which further operations on the handle fail fast with `Error::Disconnected`
+  instead of driving the dead transport.
```
jordens, line 21: Shorten this. Remove the rationalizing.

src/mqtt_client/session/tests.rs
```diff
@@ -72,22 +82,28 @@ fn session_exposes_local_packet_capacities() {
     assert_eq!(session.max_tx_packet_size(), 1152);
 }
 
+#[test]
+fn can_publish_requires_a_live_connection() {
```
jordens: That's a useless test. It only tests its own harness.
And it doesn't show the problem on dropped Connection.

src/mqtt_client/session/mod.rs
```diff
@@ -145,3 +151,130 @@ where
...
+/// [`Session::connect`](Session::connect), so correctness does not depend on this handle's `Drop`
+/// running.
+///
+/// Note that dropping or forgetting the handle is an *ungraceful* MQTT close: **no `DISCONNECT`
+/// packet is sent** (a sync `Drop` cannot perform the async write), so the broker sees an abnormal
+/// disconnect and will publish the configured Will, if any. Whether the underlying transport is
+/// actually closed is up to the `IO`'s own `Drop`. For a clean shutdown — `DISCONNECT` sent, Will
+/// suppressed — call [`disconnect`](Self::disconnect) (or [`disconnect_with`](Self::disconnect_with))
+/// before dropping. Reconnecting afterwards with [`Session::connect`](Session::connect) resumes the
+/// broker session via `CONNECT` `clean_start=false`; see [`ConnectEvent`](crate::ConnectEvent).
```
jordens: It's doesn't "release". It leaks!

All this doesn't belong here anyway.

jordens (changes requested):
I wanted verification of your premise! That motivates taking `IO` out of `Session` and separating the persistent MQTT `Session` state from the transport/network layer.
This clearly isn't just boilerplate. This is a redesign. If it was boilerplate, we wouldn't be discussing. And if you expect me to review a ~1k line PR written from an AI, you can be sure I'll have an AI do that if anything. Do not complain about getting an answer from an AI for your AI PR.

I still want the public API kept narrower than this PR: `Session` should remain MQTT state, and the live transport lifetime should be represented by an explicit live connection/lease,
not via `Deref`-magic or stale "connected" flags on `Session`.

Regarding the design I was thinking typestate `Session<'buf>::connect<IO>(self, io) -> Connection<'buf, IO>` but I suspect that will make any `Session` owner/user a bit tricky and might also make cancel-safety harder. Happy to get input on that.

Ddystopia:
Great, thanks for the review. I'll follow up around Monday.

Ddystopia:
> Regarding the design I was thinking typestate Session<'buf>::connect<IO>(self, io) -> Connection<'buf, IO> but I suspect that will make any Session owner/user a bit tricky and might also make cancel-safety harder. Happy to get input on that.

This would force every disconnect error (there are a lot of them) to return both the error and the IO, similar to `Result<(), (Session, IO, Error)>`. It will also require user to juggle the `Session` around in a loop. I personally don't think it would be worth it.

src/mqtt_client/session/mod.rs
```diff
@@ -91,15 +81,14 @@ where
         self.data.outbound.capacity()
     }
 
-    /// Return whether the session currently has the local capacity to attempt a
+    /// Return whether the session is connected and currently has the local capacity to attempt a
```
jordens:
```suggestion
    /// Return whether the session currently has the local capacity to attempt a
```

src/mqtt_client/session/operations.rs
```diff
@@ -44,32 +32,34 @@ where
...
-    /// [`Session::is_complete`](Self::is_complete), or
-    /// [`Session::is_invalidated`](Self::is_invalidated).
+    /// This is the graceful counterpart to simply dropping the handle: it sends the MQTT
+    /// `DISCONNECT` so the broker closes cleanly and suppresses the Will. Just dropping
+    /// the handle skips this and is treated by the broker as an abnormal disconnect.
+    pub async fn disconnect(&mut self) -> Result<(), Error<IO::Error>> {
+        self.disconnect_with(Disconnect::success()).await
+    }
+
+    /// Send a `SUBSCRIBE`.
```
jordens, line 52: Restore the deleted docs here (previously L57-L64).

src/mqtt_client/session/operations.rs
```diff
@@ -9,31 +9,19 @@ use crate::publication::{Publication, ToPayload};
...
-    pub async fn disconnect(&mut self) -> Result<(), Error<IO::Error>> {
-        self.disconnect_with(Disconnect::success()).await
-    }
-
-    /// Close the current MQTT transport with a caller-specified `DISCONNECT`.
-    ///
-    /// Use [`Disconnect::with_will`] to ask the broker to publish the configured Will immediately.
+impl<'buf, IO: Io> Connection<'_, 'buf, IO> {
+    /// Helper backing [`Connection::disconnect`](super::Connection::disconnect) and
+    /// [`Connection::disconnect_with`](super::Connection::disconnect_with): write the `DISCONNECT`
```
jordens: Broken refactor.
This doesn't "back disconnect_with()". It _is_ that. `super::Connection == Self`!

src/mqtt_client/session/mod.rs
```diff
@@ -145,3 +134,80 @@ where
...
+    pub fn connect_event(&self) -> ConnectEvent {
+        self.event
+    }
+
+    /// Return a reference to the underlying session.
+    pub fn session(&self) -> &Session<'buf> {
+        self.session
+    }
+
+    /// Return whether the session is connected and currently has the local capacity to attempt a
```
jordens:
```suggestion
    /// Return whether the transport is connected and the session currently has the local capacity to attempt a
```

src/mqtt_client/session/operations.rs
```diff
@@ -44,32 +32,34 @@ where
...
-    /// Call this after [`connect`](Self::connect). A resumed [`crate::ConnectEvent::Reconnected`]
-    /// already kept broker-side subscriptions.
+    /// Gracefully close the transport with `DISCONNECT`.
     ///
-    /// Returns an operation handle that can be checked with [`Session::is_pending`](Self::is_pending),
-    /// [`Session::is_complete`](Self::is_complete), or
-    /// [`Session::is_invalidated`](Self::is_invalidated).
+    /// This is the graceful counterpart to simply dropping the handle: it sends the MQTT
+    /// `DISCONNECT` so the broker closes cleanly and suppresses the Will. Just dropping
+    /// the handle skips this and is treated by the broker as an abnormal disconnect.
```
jordens, line 47:
```suggestion
    /// This is graceful and sends MQTT `DISCONNECT`. The broker suppresses the Will.
```

Otherwise it's already presented extensively in mod.rs.

jordens (changes requested):
Thanks. Only some docs tweaks.

src/mqtt_client/session/operations.rs
jordens: Ah. You fixed that in the force push.

src/mqtt_client/session/tests.rs
```diff
@@ -64,6 +64,22 @@ fn session() -> Session<'static, MockConnection> {
     )
 }
 
+/// Build a live [`Connection`] over a mock transport, bypassing the `connect` handshake, for the
+/// white-box tests below that drive the internal connection methods directly. Kept separate from
+/// `session()` so that helper still models the genuinely-disconnected `Session::new` default that
+/// the `connect`-path and pure-state tests rely on.
+fn live_connection<'a>(
```
jordens, line 71: I think you want to test Connection::can_publish

tests/async_client.rs
```diff
@@ -507,14 +493,12 @@ fn publish_requires_connected_session() {
...
-        result,
-        Err(PubError::Session(Error::Disconnected))
-    ));
-
-    expect_connected(&mut session, &connector);
-    publish_qos1_ok(&mut session, "data", b"payload");
+    // With the handle-based API, a publish is only reachable through a `Connection`,
+    // so "publish requires a connected session" is now a compile-time guarantee:
+    // there is no `Session::publish` to call before `connect`. Verify publishing
+    // works once connected.
```
jordens, line 499: But you can test `!Connection::can_publish()`

jordens (commented):
A test ensuring `!Connection::can_publish()` even for `Session::can_publish()` would be useful.

jordens (approved):

